PR #10, opened 2024-08-15
chore(deps): Switch to protobuf v5 (used in snuba)

colin-sentry:


markstory:
Looks good. I'm thinking we will end up needing a range on the generated bindings so that we don't have to upgrade several deps when protos lib updates.

PR #18, opened 2024-09-06
feat(eap): Add a "hack" message that can isolate gross stuff needed by the product

colin-sentry:
Some features like sorting on project_name can't be solved easily with our current implementation.

This adds a 'hack' message that can contain complex product logic that we need to support but don't have a good answer for yet.


volokluev (changes requested):
We do not need this, please close this PR

PR #21, opened 2024-09-09
feat(eap): add support for virtual columns in the protobuf

volokluev:
There is a requirement to have the ability to search by virtual columns. Remove the AttributeKeyTransformContext which special cased proejct id and make it generic for any desired virtual column

### DIFF
--- a/proto/sentry_protos/snuba/v1alpha/trace_item_attribute.proto
+++ b/proto/sentry_protos/snuba/v1alpha/trace_item_attribute.proto
@@ -18,9 +18,14 @@ message AttributeKey {
   string name = 2;
 }
 
-// Special cases for attributes, in particular things like project name
-message AttributeKeyTransformContext {
-  map<uint64, string> project_ids_to_names = 1;
+// custom mappings of column values which are not present in EAP storage
+// for example, project name is changeable by the user and not stored in EAP
+// but sorting by the project name may be desired. This struct carries that mapping
+// data
+message VirtualColumnContext {
+  string from_column_name = 1;
+  string to_column_name = 2;
+  map<string, string> value_map = 3;
 }
 
 message AttributeValue {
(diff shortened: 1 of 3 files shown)


colin-sentry:
how would this work for the uint case, everything gets converted to strings?

volokluev:
> how would this work for the uint case, everything gets converted to strings?

That's what I was thinking. I mostly did this because I can't do a `oneOf` map in protobuf. But if you have another solution that allows specifying both I'm happy to hear it

proto/sentry_protos/snuba/v1alpha/trace_item_attribute.proto
```diff
@@ -18,9 +18,14 @@ message AttributeKey {
...
-// Special cases for attributes, in particular things like project name
-message AttributeKeyTransformContext {
-  map<uint64, string> project_ids_to_names = 1;
+// custom mappings of column values which are not present in EAP storage
+// for example, project name is changeable by the user and not stored in EAP
+// but sorting by the project name may be desired. This struct carries that mapping
+// data
+message VirtualColumnContext {
+  string from_column_name = 1;
+  string to_column_name = 2;
```
Zylphrex, line 27: Few questions here
- Does this mean `to_column_name` will be the name of the virtual column?
- Is this name constrained anywhere or is it an arbitrary string?
- Will this guarantee the `to_column_name` will be the name of a column in the response?
- Will this name be accessible in other parts of the query? So think `count_unique(project)` will get the project names substituted in automatically? (not saying we'll do this, just want to understand the semantics of this context).

volokluev: > Does this mean to_column_name will be the name of the virtual column?

Yes 

> Is this name constrained anywhere or is it an arbitrary string?

Whatever you want bb

> Will this guarantee the to_column_name will be the name of a column in the response?

Yes

> Will this name be accessible in other parts of the query? So think count_unique(project) will get the project names substituted in automatically? (not saying we'll do this, just want to understand the semantics of this context).

This is a good question. I would say for now these virtual columns can be used to order by and in the result set. aggregations on virtual columns not supported

Zylphrex (approved):

PR #23, opened 2024-09-11
Make a dumb change

volokluev:


markstory:
This won't cause CI to fail because the modified proto is in a `v1alpha` directory. Packages where the version element ends in `alpha`, `beta` or `dev` are exempt from backwards compatibility tests.

PR #26, opened 2024-09-13
chore: Move EAP protocols out of unstable

markstory:
We're using these in code we're shipping to production. Unstable packages shouldn't be used in production systems as there are no backwards compatibility checks applied.

### DIFF
--- a/proto/sentry_protos/snuba/v1/endpoint_aggregate_bucket.proto
+++ b/proto/sentry_protos/snuba/v1/endpoint_aggregate_bucket.proto
@@ -1,10 +1,10 @@
 syntax = "proto3";
 
-package sentry_protos.snuba.v1alpha;
+package sentry_protos.snuba.v1;
 
-import "sentry_protos/snuba/v1alpha/request_common.proto";
-import "sentry_protos/snuba/v1alpha/trace_item_attribute.proto";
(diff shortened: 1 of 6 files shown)


volokluev:
The components that we are using this for are not serving any customers. Has there been an issue that was caused by backwards incompatibility?

volokluev:
These APIs are 100% not ready to be backwards compatible yet. We're learning new things every day about how they're supposed to work

markstory:
> The components that we are using this for are not serving any customers. 

That's good to hear 👍 

> Has there been an issue that was caused by backwards incompatibility?

The issue I'm aware of is that bumping sentry-protos caused CI failures due to breaking changes. The concern I have is how do we ensure that alpha protos don't creep into workloads that are handling customer traffic accidentally. One simple solution I can think of is to not include unstable protos in release packages, but that impacts CI while systems are being developed.

markstory:
Not going to merge this until we come to a decision on how we want to handle unstable protos more. See #27 27

volokluev:
@markstory I think this PR will actually break CI again. I think the better thing to do would be to copy this folder into V1. I have some changes I want to make to the interface anyways so I can take this on. For now I am in agreement that no more backwards incompatible changes will be made on this interface.

PR #27, opened 2024-09-13
feat: Exclude unstable protos in release builds

markstory:
Including unstable protos in release builds opens the opportunity that we ship a feature with unstable protos and create a breaking change.

By excluding unstable protos from release builds we prevent this scenario from happening. However, it creates some undesirable side-effects:

1. Code relying on unstable protos can't easily be checked in to application repositories. This limits teams ability to collaborate on functionality that relies on unstable protos.
2. Creates 'works on my machine' scenarios where a local build of proto stubs is different than what CI and production will use.

Is this a process we want to have? If we don't go with this solution how else can we prevent unstable protos from being used with production traffic?


corps:
> Is this a process we want to have? If we don't go with this solution how else can we prevent unstable protos from being used with production traffic?

Linters?  I don't know if it's the right solution, but it is one: say that main branches of repos should lint against imports of unstable protos.

But preventing unstable protos it from being packaged into releases is reasonable since someone can still produce a wheel manually and check that into their branch if they so desire.

What if we just introduce a makefile target that builds a local wheel and outputs instructions on how to use that on a branch?

volokluev:
I just want to describe how we're actually using this repo on EAP right now:

`sentry-protos` is the repo we use to define the interface between two services. We definitely get the interface wrong and have had to change it multiple times. 

The code that uses these protos *does* run in production behind a feature flag and it's okay if it breaks. 


I'm not interested in a solution that allows me to make breaking changes but only locally, that will not help. 

I am interested in specifics of how this caused CI failures so we can come up with something that will help

asottile-sentry:
> I just want to describe how we're actually using this repo on EAP right now:
> 
> `sentry-protos` is the repo we use to define the interface between two services. We definitely get the interface wrong and have had to change it multiple times.
> 
> The code that uses these protos _does_ run in production behind a feature flag and it's okay if it breaks.
> 
> I'm not interested in a solution that allows me to make breaking changes but only locally, that will not help.
> 
> I am interested in specifics of how this caused CI failures so we can come up with something that will help

the breaking changes made `sentry-protos` unbumpable: https://github.com/getsentry/sentry/actions/runs/10850857555/job/30113281965

volokluev:
Thank you @asottile-sentry for the context.

I see the problem now which is that the unstable changes are coupled with stable ones and that makes our workflow impossible without breaking stuff. 

I think enforcing the backwards compatibility requirement is fine and for usecases like this, we should just be really liberal with major version bumps if we don't actually care about backwards compatibility. That's what I will instruct the team to do going forward.

Sound good?

asottile-sentry:
> Thank you @asottile-sentry for the context.
> 
> I see the problem now which is that the unstable changes are coupled with stable ones and that makes our workflow impossible without breaking stuff.
> 
> I think enforcing the backwards compatibility requirement is fine and for usecases like this, we should just be really liberal with major version bumps if we don't actually care about backwards compatibility. That's what I will instruct the team to do going forward.
> 
> Sound good?

yep versions are free -- fwiw this is what we decided on in the original protobufs meeting as well

markstory:
> What if we just introduce a makefile target that builds a local wheel and outputs instructions on how to use that on a branch?

We have those with these changes. `make build-py` and `make build-rust` will generate unstable protos for use in local development.

> I think enforcing the backwards compatibility requirement is fine and for usecases like this, we should just be really liberal with major version bumps if we don't actually care about backwards compatibility. That's what I will instruct the team to do going forward.

That works for me too. Like Anthony said, version numbers are free, and if we get to v20 on protos that's ok.

volokluev:
I think eventually we should start thinking about tests for this code we are writing

Makefile
```diff
@@ -1,6 +1,9 @@
 # Makefile assumes that direnv is active, or that pip/python on PATH
 # is what you want to use.
 
+# unstable protos are only included in local development and not part of release packages
+SENTRY_PROTOS_BUILD_UNSTABLE := 1
```
corps, line 5: I'm a little confused.  Won't this always activate?  We also use these make targets in CI so... https://github.com/getsentry/sentry-protos/blob/main/.github/workflows/build.yml#L23

Maybe I'm misremembering makefile precedence, but environment variables set in this way override ones provided by command line or environment?  So this would be 100%

corps (approved):

markstory:
> I think eventually we should start thinking about tests for this code we are writing

There are 'tests' to make sure that codegen is working. I would like to expand the tests more, but also don't want to have 'junk' protos around to validate unstable build behavior. 🤔

Makefile
markstory: The `package-py` target sets the environment variable value when it runs `build-py`. Environment variables also overwrite the default values defined in `Makefile`

corps:
> > I think eventually we should start thinking about tests for this code we are writing
> 
> There are 'tests' to make sure that codegen is working. I would like to expand the tests more, but also don't want to have 'junk' protos around to validate unstable build behavior. 🤔

In all fairness, the official repo does have "junk" test protos lying around :P 
If we prefix it with an unstable version, and it doesn't build into packaging, would that be reasonable enough compromise?

markstory:
> If we prefix it with an unstable version, and it doesn't build into packaging, would that be reasonable enough compromise?

That could work. Tests could build unstable and stable protos and make assertions on the built output.

PR #28, opened 2024-09-13
feat(eap): Qu-EAP V1

volokluev:
# Context

https://www.notion.so/sentry/Qu-EAP-Querying-Events-Analytics-Platform-76629acc212b4ac486c63e7979f16076#c6dc0f7df06849bc808b5989a2e7c1bb

This PR is the first interface proposal for Qu-EAP, coming from the alpha version where we experimented a lot 

This PR copies the `alpha` protos into `v1` and makes some changes to the API interface to keep it in line with the EAP vision. Namely:

* Rename the `span_samples` endpoint to `trace_item_columns` 
    * EAP was never meant to only serve spans, naming the endpoint in this way is going against the vision
    * This endpoint is not in use by anything right now so it's easier to rename now
* Make the `TraceItemColumnResponse` no longer re-encode the column name for every value
* Add `page_token` to `TraceItemColumnRequest`, `TraceItemColumnResponse`, `AttributeValuesRequest`, `AttributeValuesResponse`
    * Historically, we have used limit and offset to do pagination which is inefficient because all the scanning and ordering still needs to happen in order to use these. By allowing a `page_token` parameter, we can be more flexible with how we actually do pagination and allow us to actually narrow down search results
    
# Desired Feedback

* The interface is not missing an important feature that is necessary for the performance product
* Naming is understandable and consistent
* The purpose of every endpoint is clear
* The README gives sufficient context to understand the repository


proto/sentry_protos/snuba/v1/endpoint_aggregate_bucket.proto
```diff
@@ -0,0 +1,33 @@
+syntax = "proto3";
```
Zylphrex, line 1: Due to the TODO, we will have breaking api changes to this rpc to support multiple y axis and group bys. Not sure what else yet but just wanted to call this out.

proto/sentry_protos/snuba/v1/endpoint_tags_list.proto
```diff
@@ -0,0 +1,37 @@
...
+
+message AttributeValuesRequest {
+  RequestMeta meta = 1;
+  string name = 3;
+  // a substring of the value being searched,
+  // only strict substring supported, no regex
+  string value_substring_match = 4;
+  uint32 limit = 5;
+  string page_token = 6;
+}
```
Zylphrex, line 32: One question I had while working on this is if this will support non attribute columns like I see segment name is available. Because autocompletion those would be nice to have. In which case it may be smart to turn the name into a attribute key

proto/sentry_protos/snuba/v1/endpoint_tags_list.proto
```diff
@@ -0,0 +1,37 @@
...
+
+message AttributeValuesResponse {
+  repeated string values = 1;
```
Zylphrex, line 35: Some nice to haves per value that I found while working on this are times seen/first seen/last seen. Not necessary but as I'm not confident what it's used for in the UI yet but just wanted to raise it.

proto/sentry_protos/snuba/v1/endpoint_aggregate_bucket.proto
```diff
@@ -0,0 +1,33 @@
...
+    FUNCTION_P95 = 5;
+    FUNCTION_P99 = 6;
+    FUNCTION_AVG = 7;
+  }
+
+  Function aggregate = 4;
+  TraceItemFilter filter = 5;
+  uint64 granularity_secs = 6;
+  AttributeKey key = 7;
+  repeated VirtualColumnContext virtual_column_context = 8;
```
Zylphrex, line 26: Not sure how this will be used here yet. Is it meant for the top n use case?

proto/sentry_protos/snuba/v1/endpoint_aggregate_bucket.proto
volokluev: Added

proto/sentry_protos/snuba/v1/endpoint_tags_list.proto
volokluev: Good question, in EAP we don't distinguish between columns which are dynamic vs directly on the table so yes it should return both

proto/sentry_protos/snuba/v1/endpoint_aggregate_bucket.proto
volokluev: I'm happy to remove it for now

proto/sentry_protos/snuba/v1/trace_item_attribute.proto
```diff
@@ -0,0 +1,38 @@
...
+// for example, project name is changeable by the user and not stored in EAP
+// but sorting by the project name may be desired. This struct carries that mapping
+// data
+message VirtualColumnContext {
+  string from_column_name = 1;
+  string to_column_name = 2;
+  map<string, string> value_map = 3;
+}
+
+message AttributeValue {
```
vgrozdanic, line 31: Do we expect the `AttributeValue` ever to be `null`?
I am not very familiar with protobufs, but we might need to use wrappers or optional keyword here if we want to have null as a value?

proto/sentry_protos/snuba/README.md
```diff
@@ -0,0 +1,47 @@
...
+    `project_id` to name mapping is passed with the query so that project_name can be a virtual column
+
+
+# Is this Protobuf definition a re-implementation of [SnQL](https://getsentry.github.io/snuba/language/snql.html)?
+
+No, Qu-EAP differes from SnQL in key ways:
+
+1. Qu-EAP does not distinguish between data stored as a column or a tag, TraceItems have attributes and the interface to query them is all the same.
+2. Qu-EAP operations are limited. Currently there are four:
+    1. A timessries request (with a predefined list of possible aggregations)
```
vgrozdanic, line 32: nit: a typo "timessries" -> "timeseries"

proto/sentry_protos/snuba/v1/trace_item_filter.proto
```diff
@@ -0,0 +1,45 @@
...
+  AttributeValue value = 3;
+}
+
+message ExistsFilter {
+  AttributeKey key = 1;
+}
+
+// Represents a condition on searching for a particular "trace item"
+// (e.g., spans, replays, errors)
+message TraceItemFilter {
```
vgrozdanic, line 38: Do we want to support `NotFilter` to negate the filter condition?

proto/sentry_protos/snuba/v1/endpoint_aggregate_bucket.proto
```diff
@@ -0,0 +1,44 @@
+syntax = "proto3";
+
+package sentry_protos.snuba.v1;
+
+import "sentry_protos/snuba/v1/request_common.proto";
+import "sentry_protos/snuba/v1/trace_item_attribute.proto";
+import "sentry_protos/snuba/v1/trace_item_filter.proto";
+
+enum Function {
```
vgrozdanic, line 9: Are the `min`, `max`, `p90` and `uniq` [left out](https://www.notion.so/sentry/Qu-EAP-Querying-Events-Analytics-Platform-76629acc212b4ac486c63e7979f16076?pvs=4#817970de59f7455093457621a621dd19) on purpose?

proto/sentry_protos/snuba/v1/request_common.proto
```diff
@@ -0,0 +1,32 @@
...
+  google.protobuf.Timestamp end_timestamp = 6;
+  TraceItemName trace_item_name = 7;
+}
+
+enum TraceItemName {
+  TRACE_ITEM_NAME_UNSPECIFIED = 0;
+  TRACE_ITEM_NAME_EAP_SPANS = 1;
+}
+
+message PageToken {
```
volokluev, line 24: <details>

<summary>Conversation Colin and I had about this (for historical posterity) </summary>

Colin:
does clickhouse work well with ID-based pagination? wouldn't it be hard to convert a span id to an offset to use in a subquery?
because "find span by id" is slow in current config


Volo Kluev
ah yes let's chat about that, would be good to question my assumptions here
You are right that ID-based pagination would not work
What I'm imagining is a filter that is passed back by the server that is based on the order by.
https://github.com/getsentry/sentry-protos/pull/28/files#diff-94d63309faaad438fa788b21a8031e155ff059efeec32e921996e39257226bb7R27-R30
So for example, if in the first request you ordered by (timestamp, project_name), the server would take the max timestamp and project name and pass that back in the page token


Colin Chartier
you could maybe use (timestamp, id) pairs, but that would be weird devex


Volo Kluev
the devex I'm imagining is that the server always passes back a pagetoken and all the client has to do is pass it in to paginate


Colin Chartier
could we just return like a hash string
struct pack (timestamp, span id) and then hex encode that
that's how like reddit or github pagination works

Volo Kluev
yes we could, I was thinking of that at first but that is essentially what protobuf is doing, it's binary encoding the conditions of the filter


Colin Chartier
the trace item filter one is clever, but where would it go
into a subquery?
because if your filter is (span_id = blah and timestamp = blah)
then you can't just AND that with the existing query, you need to do OFFSET (select count() from eap_spans where span_id = blah and timestamp = blah)
or something equivalent
if the condition is only on timestamp, then why not just send the timestamp instead of the whole structure

Volo Kluev
the filter would be something like (timestamp > X) and I think I could AND that with the existing query


Colin Chartier
the alternative is you can just have a list of keys, and just do like
where (org_id, project_id, _sort_timestamp, span_id) > (a, b, c, d)
which would hit the primary key
but that sort of condition can't currently be encoded by traceitemfilter
i think of these options, the one I like best is just
repeated TraceItemKey cursor = 1
or even repeated string cursor = 1


Volo Kluev
What about
(org_id > a AND project_id > b AND _sort_timestamp > c AND span_id > d)


Colin Chartier
that could work, yeah
ok, i'm fine with that and encoding it as traceitemfilter too

proto/sentry_protos/snuba/v1/request_common.proto
```diff
@@ -0,0 +1,32 @@
...
+
+message RequestMeta {
+  uint64 organization_id = 1;
+  string cogs_category = 2;
+  string referrer = 3;
+  // can be empty
+  repeated uint64 project_ids = 4;
+  google.protobuf.Timestamp start_timestamp = 5;
+  google.protobuf.Timestamp end_timestamp = 6;
+  TraceItemName trace_item_name = 7;
```
colin-sentry, line 16: should probably be TraceItemType

proto/sentry_protos/snuba/v1/trace_item_attribute.proto
```diff
@@ -0,0 +1,38 @@
+syntax = "proto3";
+
+package sentry_protos.snuba.v1;
+
+message AttributeKey {
+  enum Type { //this mostly reflects what types are able to be ingested, see eap_spans consumer for ingest details
+    TYPE_UNSPECIFIED = 0; //called "none" sometimes
```
colin-sentry, line 7: isn't "none" in the context of units? would we ever do anything but error if someone asked for an untyped attribute?

proto/sentry_protos/snuba/v1/trace_item_attribute.proto
```diff
@@ -0,0 +1,38 @@
...
+    TYPE_BOOLEAN = 1;
+    TYPE_STRING = 2;
+    TYPE_FLOAT = 3;
+    TYPE_INT = 4; //note: all numbers are stored as float64, so massive integers can be rounded. USE STRING FOR IDS.
+  }
+
+  Type type = 1;
+
+  //if you use span_id this will route to span.span_id,
```
colin-sentry, line 16: comment should be updated RE the namespacing, "if you use sentry.span_id, this will route to span.span_id, if you use derp this will route to ..."

proto/sentry_protos/snuba/v1/endpoint_aggregate_bucket.proto
```diff
@@ -0,0 +1,47 @@
...
+}
+
+message TimeSeries {
+  string label = 1;
+  // the names of the attribute keys which
+  // were in the group by
+  repeated string attribute_names = 2;
+  repeated float data_points = 3;
+  // number of events used to calculate this timeseries
+  uint64 num_events = 4;
```
colin-sentry, line 41: @Zylphrex and @wmak - is this the querying interface you want for timeseries? I still am not completely sure how you want this RPC to work

colin-sentry: I think you said you wanted granularity and interval for the request instead?

proto/sentry_protos/snuba/v1/request_common.proto
volokluev: Why is that? I imagine a `type` to signify a different category which this does not. For example, issues and spans are not different types, they are both trace items, they are accessed with the same RPC calls, but they have different names.

proto/sentry_protos/snuba/v1/trace_item_attribute.proto
volokluev: This is a carryover from V0, I'll remove it

proto/sentry_protos/snuba/v1/trace_item_attribute.proto
volokluev: Would `duration` now be `sentry.duration`?

colin-sentry: yes, we're namespacing everything that isn't user sent to `sentry.` - that way the sentry backend can choose how to alias things if there are conflicts

proto/sentry_protos/snuba/v1/trace_item_filter.proto
volokluev: Good point, I shall add

proto/sentry_protos/snuba/v1/trace_item_attribute.proto
volokluev: Good point

proto/sentry_protos/snuba/v1/endpoint_aggregate_bucket.proto
```diff
@@ -0,0 +1,47 @@
...
+  FUNCTION_P99 = 6;
+  FUNCTION_AVG = 7;
+}
+
+message AggregateBucketRequest {
+  RequestMeta meta = 1;
+  TraceItemFilter filter = 2;
+  repeated AttributeAggregation aggregations = 3;
+  uint64 granularity_secs = 4;
+  repeated AttributeKey group_by = 5;
```
wmak, line 29: Is there a way to limit the number of results the group by will cause? As well we'll likely need an orderby too

proto/sentry_protos/snuba/v1/endpoint_aggregate_bucket.proto
```diff
@@ -0,0 +1,47 @@
...
+  Function aggregate = 1;
+  AttributeKey key = 2;
+  string label = 3;
+}
+
+message TimeSeries {
+  string label = 1;
+  // the names of the attribute keys which
+  // were in the group by
+  repeated string attribute_names = 2;
```
wmak, line 42: Should this be `attribute_values`? Like we query for `p50()` grouped by `transaction` and `organization` this will tell us that for this timeseries its for the `/api` transaction in the `sentry` org?

proto/sentry_protos/snuba/v1/endpoint_aggregate_bucket.proto
```diff
@@ -0,0 +1,47 @@
...
+}
+
+message AttributeAggregation {
+  Function aggregate = 1;
+  AttributeKey key = 2;
+  string label = 3;
+}
+
+message TimeSeries {
+  string label = 1;
```
wmak, line 39: What is Label?

proto/sentry_protos/snuba/v1/endpoint_aggregate_bucket.proto
evanh: Why would the query be ordered by anything other than time?

proto/sentry_protos/snuba/README.md
```diff
@@ -0,0 +1,48 @@
...
+    `project_name` is not
+    `project_id` to name mapping is passed with the query so that project_name can be a virtual column
+
+
+# Is this Protobuf definition a re-implementation of [SnQL](https://getsentry.github.io/snuba/language/snql.html)?
+
+No, Qu-EAP differes from SnQL in key ways:
+
+1. Qu-EAP does not distinguish between data stored as a column or a tag, TraceItems have attributes and the interface to query them is all the same.
+    1. Note: attributes sent by sentry are prefixed with `sentry.{attr_name}` to allow users to send custom tags with the same name.
```
wmak, line 31: Should we do the reverse as well, prefix user tags with `user` or something similar? That way users can't send tags that overlap sentry tags by sending `sentry.blah`, it'll become `user.sentry.blah`

proto/sentry_protos/snuba/v1/endpoint_aggregate_bucket.proto
```diff
@@ -0,0 +1,51 @@
...
+  repeated AttributeKey group_by = 5;
+}
+
+message AttributeAggregation {
+  Function aggregate = 1;
+  AttributeKey key = 2;
+  string label = 3;
+}
+
+message TimeSeries {
```
wmak, line 38: Actually to be sure this is the `TimeSeriesResponse` right? Should we maybe name it that to match the other protos?

proto/sentry_protos/snuba/v1/endpoint_trace_item_attributes.proto
```diff
@@ -0,0 +1,44 @@
...
+package sentry_protos.snuba.v1;
+
+import "sentry_protos/snuba/v1/request_common.proto";
+import "sentry_protos/snuba/v1/trace_item_attribute.proto";
+
+//A request for "which queryable attributes are available for these projects between these dates"
+// - used for things like autocompletion
+// EAP does not make a distinction between data stored as columns on a table vs attributes stored
+// dynamically (e.g. tags, contexts, etc). To the user of the interface, they are the same
+message TraceItemAttributesRequest {
```
wmak, line 12: The naming of this and `proto/sentry_protos/snuba/v1/endpoint_trace_item_columns.proto`
confuse me a bit, so might be misunderstanding

Here's my current guess;

TraceItemAttributesRequest -> Gets me the names of attrs, excluding columns?
TraceItemColumnsRequest -> Gets the columns, excluding attrs?
AttributeValuesRequest -> Gets me the values for a given attr or column

proto/sentry_protos/snuba/v1/endpoint_trace_item_attributes.proto
```diff
@@ -0,0 +1,44 @@
...
+  message Attribute {
+    string name = 1;
+    AttributeKey.Type type = 2;
+  }
+  repeated Attribute attributes = 1;
+}
+
+// This endpoint only supports string values, it does not make sense
+// to get all the possible values of a numerical attribute
+message AttributeValuesRequest {
```
wmak, line 31: If my previous guess is correct should we rename this to `TraceItemAttributeValuesRequest` and the other one to `TraceItemAttributeNamesRequest`?

proto/sentry_protos/snuba/v1/trace_item_filter.proto
```diff
@@ -0,0 +1,50 @@
...
+  enum Op {
+    OP_UNSPECIFIED = 0;
+    OP_LESS_THAN = 1;
+    OP_GREATER_THAN = 2;
+    OP_LESS_THAN_OR_EQUALS = 3;
+    OP_GREATER_THAN_OR_EQUALS = 4;
+    OP_EQUALS = 5;
+    OP_NOT_EQUALS = 6;
+    OP_LIKE = 7; //string only
+    OP_NOT_LIKE = 8; //string only
```
wmak, line 29: can we add `OP_IN` and `OP_NOT_IN?`

proto/sentry_protos/snuba/v1/endpoint_aggregate_bucket.proto
```diff
@@ -0,0 +1,51 @@
...
+  RequestMeta meta = 1;
+  TraceItemFilter filter = 2;
+  repeated AttributeAggregation aggregations = 3;
+  uint64 granularity_secs = 4;
+  repeated AttributeKey group_by = 5;
+}
+
+message AttributeAggregation {
+  Function aggregate = 1;
+  AttributeKey key = 2;
```
wmak, line 34: Should we try to make this more future proof by allowing multiple keys?

proto/sentry_protos/snuba/v1/endpoint_aggregate_bucket.proto
```diff
@@ -0,0 +1,51 @@
...
+  string label = 3;
```
evanh, line 35: What is this for?

wmak (commented):
Hmm would be helpful in reviewing this to see examples instead of having to read just the proto definitions if possible

proto/sentry_protos/snuba/v1/endpoint_aggregate_bucket.proto
```diff
@@ -0,0 +1,51 @@
...
+}
+
+message TimeSeries {
+  string label = 1;
+  // the names of the attribute keys which
+  // were in the group by
+  repeated string attribute_names = 2;
```
evanh, line 42: I am a little confused about how this works with group by.

In theory there would be one timeseries for each unique combination of attribute values across all the attribute keys in the GROUP BY.

Does this field correspond to those values? And if so, how can these values be tied back to the corresponding keys?

proto/sentry_protos/snuba/v1/endpoint_trace_item_attributes.proto
evanh, line 12: I think this should have the `value_substring_match` field as well, for auto complete.

proto/sentry_protos/snuba/v1/endpoint_trace_item_columns.proto
```diff
@@ -0,0 +1,32 @@
+syntax = "proto3";
+
+package sentry_protos.snuba.v1;
+
+import "sentry_protos/snuba/v1/request_common.proto";
+import "sentry_protos/snuba/v1/trace_item_attribute.proto";
+import "sentry_protos/snuba/v1/trace_item_filter.proto";
+
+message TraceItemColumnsRequest {
```
evanh, line 9: What is this endpoint for?

proto/sentry_protos/snuba/v1/endpoint_trace_item_attributes.proto
volokluev: > TraceItemAttributeNamesRequest

Yes much better name!

proto/sentry_protos/snuba/v1/endpoint_trace_item_attributes.proto
volokluev: I can see how this is confusing. I renamed 

`TraceItemAttributesRequest -> TraceItemAttributeNamesRequest`
`AttributeValuesRequest -> TraceItemAttributeValuesRequest`

For  `TraceItemAttributesRequest` I could use some help. This is the endpoint that is intended to serve tabular queries. Like your more standard `SELECT project_id, duration, span_name FROM spans` hence I named it columns. I'm open to other ideas of how to name this

proto/sentry_protos/snuba/v1/endpoint_aggregate_bucket.proto
volokluev: I renamed the file to make everything based on `TimeSeries`. The `TimeSeriesResponse` consists of many `TimeSeries`

proto/sentry_protos/snuba/v1/endpoint_aggregate_bucket.proto
volokluev: The endpoint allows for multiple `AttributeAggregation`s

proto/sentry_protos/snuba/v1/endpoint_trace_item_attributes.proto
volokluev: Done

proto/sentry_protos/snuba/v1/endpoint_trace_item_columns.proto
volokluev: This endpoint is to serve table views

proto/sentry_protos/snuba/README.md
colin-sentry: that should presumably be blocked by sdks and in relay

colin-sentry: (it's against otel spec to use someone else's namespace)

wmak: Good thing user's will read the entirety of otel spec then right ;)

colin-sentry: eh, ideally the SDKs can just tell them to knock it off

proto/sentry_protos/snuba/v1/trace_item_filter.proto
volokluev: Done

colin-sentry: we'd need a new class for this, right? since `value` here isn't a list

proto/sentry_protos/snuba/v1/endpoint_trace_item_columns.proto
wmak: Would naming it `TraceItemTableRequest` perhaps be better? Might not be as abstract but communicates the intent better

proto/sentry_protos/snuba/v1/endpoint_time_series.proto
```diff
@@ -0,0 +1,51 @@
...
+  FUNCTION_MIN = 10;
+  FUNCTION_UNIQ = 11;
+}
+
+message TimeSeriesRequest {
+  RequestMeta meta = 1;
+  TraceItemFilter filter = 2;
+  repeated AttributeAggregation aggregations = 3;
+  uint64 granularity_secs = 4;
+  repeated AttributeKey group_by = 5;
```
colin-sentry, line 29: what happens if we group by something with high cardinality? sentry itself has this "top n" functionality that makes it so it only returns at most like 10 groups with the remainder put in "other", is that going to be the behavior here?

colin-sentry: cc @wmak

wmak: > Why would the query be ordered by anything other than time?
cc @evanh 

I dunno where the comment went, but this is why I'm asking for `order by` and `limit by` so we can use it with `group by`

proto/sentry_protos/snuba/v1/endpoint_time_series.proto
```diff
@@ -0,0 +1,51 @@
...
+  Function aggregate = 1;
+  AttributeKey key = 2;
+  string label = 3;
+}
+
+message TimeSeries {
+  string label = 1;
+  // the names of the attribute keys which
+  // were in the group by
+  repeated string attribute_names = 2;
```
colin-sentry, line 42: i'm not sure about this - what happens if you specify multiple aggregations and a group by, wouldn't we need a map<string, repeated float> where the map is for the group by, and the repeated float is one per aggregation?

proto/sentry_protos/snuba/v1/endpoint_time_series.proto
```diff
@@ -0,0 +1,51 @@
...
+  repeated float data_points = 3;
+  // number of events used to calculate this timeseries
+  uint64 num_events = 4;
+  float avg_sampling_rate = 5;
```
colin-sentry, line 46: why include these? they are already encompassed by adding an aggregate with `count` as the function, for example

proto/sentry_protos/snuba/README.md
wmak: 🤔 what's the interface or experience you're imagining to tell them to knock it off?

colin-sentry: if you `span.setData("sentry.haha", "blah")` it says "you are not allowed to do that."

wmak: 🤔 via a warning? error? print statement?
does the data still get set? or do we null that data?
if `haha` is a key we rely on does `setData` still work? or becomes a no-op?

proto/sentry_protos/snuba/v1/endpoint_trace_item_columns.proto
```diff
@@ -0,0 +1,32 @@
...
+  message OrderBy {
+    AttributeKey key = 1;
+    bool descending = 2;
+  }
+
+  RequestMeta meta = 1;
+  TraceItemFilter filter = 2;
+  repeated OrderBy order_by = 3;
+  repeated AttributeKey keys = 4;
```
evanh, line 18: Is this just a table of raw attribute values? Is there a way to have aggregates in the table?

volokluev: Correct and there are no aggregates, although maybe that's an oversight

proto/sentry_protos/snuba/v1/endpoint_aggregate_bucket.proto
volokluev: yes you are correct, this needs to be a map<str, str> for attribute names and values

proto/sentry_protos/snuba/v1/examples.py
```diff
@@ -0,0 +1,96 @@
...
+    organization_id=1,
+    cogs_category="something",
+    referrer="something",
+    start_timestamp=Timestamp(seconds=int(datetime(2024, 4, 20, 16, 20).timestamp())),
+    end_timestamp=Timestamp(seconds=int(datetime(2024, 4, 20, 17, 20).timestamp())),
+    trace_item_name=TraceItemName.TRACE_ITEM_NAME_EAP_SPANS,
+)
+
+
+def test_example_time_series():
```
volokluev, line 31: @wmak thank you for the prompt to add examples. I will start adding them here as I do them. Also invite you to try and do your own as well and see if the mental model breaks anywhere

proto/sentry_protos/snuba/v1/examples.py
```diff
@@ -0,0 +1,96 @@
...
+            ),
+        ],
+        granularity_secs=60,
+        group_by=[
+            AttributeKey(type=AttributeKey.TYPE_STRING, name="endpoint_name"),
+            AttributeKey(type=AttributeKey.TYPE_STRING, name="consumer_group"),
+        ],
+    )
+
+    TimeSeriesResponse(
```
wmak, line 53: This is already helping me 🙏 

So for the timeseries, is it the callers responsibility to determine what the buckets were right? Or is that something we can include in the response?

And by bucket i mean `data_points[0]` happening at `2024-4-20 16:20:00` and then figuring out that `data_points[1]` etc. happened 60s later based on `granularity_secs`

proto/sentry_protos/snuba/v1/examples.py
```diff
@@ -0,0 +1,96 @@
...
+        result_timeseries=[
+            TimeSeries(
+                label="p50",
+                group_by_attributes={
+                    "endpoint_name": "/v1/rpc",
+                    "consumer_group": "snuba_outcomes_consumer",
+                },
+                data_points=[42 for _ in range(60)],
+                num_events=1337,
+                avg_sampling_rate=0.1,
```
wmak, line 63: I'm still working out what extrapolation will look like, but my gut feeling is that we'll need sampling_rate for each data point

proto/sentry_protos/snuba/v1/examples.py
volokluev: we can definitely include the buckets!

proto/sentry_protos/snuba/v1/endpoint_trace_item_columns.proto
volokluev: Sounds good

proto/sentry_protos/snuba/v1/endpoint_time_series.proto
wmak: 👍 that's fine with me, so to match the current behaviour today then we'll keep doing what we've already been doing which is;

1. make a `TraceItemTableRequest` to get the totals
2. Pass those totals as part of the filter here to `TimeSeriesRequest`
3. Make a parallel query with the inverse filter for the `other`

In which case just the group-by should be fine

proto/sentry_protos/snuba/v1/endpoint_trace_item_columns.proto
volokluev: added aggregates

proto/sentry_protos/snuba/v1/endpoint_time_series.proto
volokluev: https://github.com/getsentry/eap-planning/issues/35

proto/sentry_protos/snuba/v1/endpoint_time_series.proto
volokluev: This should be addressed now, see examples

proto/sentry_protos/snuba/v1/endpoint_aggregate_bucket.proto
volokluev: It's basically an alias, I didn't want to use the same word because it's not SQL but maybe alias is less confusing. What do you think?

proto/sentry_protos/snuba/v1/endpoint_aggregate_bucket.proto
volokluev: basically an alias. It's optional. Should I just use `alias` instead?

wmak: 🤔 yeah, alias is the term already in use so would be easier to grok if we just used it consistently

volokluev:
We've had a lot of good discussion on this PR:

I've added examples for the table and timeserieis endpoints. What would yall need to move approve this?

@Zylphrex @wmak @evanh @colin-sentry ?

colin-sentry (approved):

proto/sentry_protos/snuba/README.md
evanh: Is there some reason not to namespace user data?

proto/sentry_protos/snuba/v1/endpoint_aggregate_bucket.proto
evanh: I would stick to alias. It makes more sense to me in this context.

proto/sentry_protos/snuba/v1/endpoint_time_series.proto
```diff
@@ -0,0 +1,33 @@
...
+  repeated AttributeKey group_by = 5;
+}
+
+message TimeSeries {
+  string label = 1;
+  // the names and values of the attribute keys which
+  // were in the group by
+  map<string, string> group_by_attributes = 2;
+  // time bucket for each data point in the timeseries
+  repeated google.protobuf.Timestamp buckets = 3;
```
evanh, line 24: Are you enforcing somewhere that every timeseries has a data point for every bucket? Otherwise you can get sparse time series that don't have data for specific buckets.

evanh (approved):

PR #29, opened 2024-09-16
Support vendored google protobuf sources.

corps:
Generated files are distributed with protobuf libraries, so they do not need to be built even if they are referenced.

Follow up PR creates dependency on `google.protobuf.struct`


buf.yaml
```diff
@@ -1,11 +1,11 @@
 version: v2
 modules:
   - path: proto
+    excludes:
+      - proto/google/protobuf
```
corps, line 5: No need to actually build these files in practice, as the generated message types are distributed as part of our python runtime anyways.

buf.yaml
```diff
@@ -1,11 +1,11 @@
...
 breaking:
   use:
     - FILE
   ignore_unstable_packages: true
-lint:
   ignore:
-    - "kafka/events.v1.proto"
-    - "relay/event.proto"
+    - google/proto
```
corps, line 11: Also no need to lint these files.

Makefile
```diff
@@ -30,5 +30,18 @@ build-rust:
 clean-rust:
 	cd rust && cargo clean
 
+repodir := $(dir $($(abspath $(lastword $(MAKEFILE_LIST)))))
+.PHONY: update-vendor
+update-vendor:
+	cd $$(mktemp -d) && \
```
corps, line 36: Vendoring is just a shallow checkout and directory replacement.

Makefile
```diff
@@ -30,5 +30,18 @@ build-rust:
...
+		git clone -n --depth=1 --filter=tree:0 \
+		    https://github.com/protocolbuffers/protobuf && \
+		cd protobuf && git sparse-checkout set --no-cone src/google/protobuf && \
+		git checkout && rm -rf .git && \
+		find . '(' ! -name '*.proto' -a ! -name '*.md' ')' -delete && \
+		find . && \
+		rm -rf $(repodir)proto/google && \
+		mv src/google $(repodir)proto/
```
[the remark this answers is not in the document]
corps: Fair!  I'll keep pruning then.

markstory (approved):
Looks good to me.

PR #30, opened 2024-09-16
Adds seer summary endpoints

corps:
Starting with the summary endpoints, implements https://github.com/getsentry/seer/blob/main/src/seer/automation/summarize/models.py#L8


markstory (approved):
Looks good to me as far as proto definitions go.

PR #34, opened 2024-09-23
feat(taskworker) Move and expand protos for taskworker

markstory:
- Move protos for taskworker into a more conventional location.
- Expand and rename protos for WorkerTask
- Add service RPC methods.

Refs getsentry/sentry#77948


proto/sentry_protos/sentry/v1alpha/taskworker.proto
```diff
@@ -0,0 +1,141 @@
...
+// RPC messages and services
+////////////////////////////
+
+// RPC interface for TaskConsumers.
+// These methods would be used in pull based models.
+service TaskConsumerService {
+  // Get a task to process
+  rpc GetTask(GetTaskRequest) returns (GetTaskResponse);
+  // Update the state of a task with execution results. (also used in push models)
+  rpc CompleteTask(CompleteTaskRequest) returns (CompleteTaskResponse);
```
enochtangg, line 109: Since this rpc takes any input `TaskState`. Should we call it `SetTask` instead of `CompleteTask`? This makes it clear that the method is meant to update the task state to any of the enum values.

proto/sentry_protos/sentry/v1alpha/taskworker.proto
```diff
@@ -0,0 +1,141 @@
...
+}
+
+message GetTaskRequest {}
+message GetTaskResponse {
+  WorkerTask task = 1;
```
enochtangg, line 114: In the pull model prototype that we created last week, we had a similar `GetTaskResponse` schema as you defined here. Consider what happens when there are no pending messages available in the pending task store. In this case, the GRPC server will return a `GetTaskResponse` where `.task` is "empty". In the GRPC client, we noticed that `.task` is not None but instead type `WorkerTask` where each of its attributes are not set. As a result, we used `.HasField` to determine whether or not task is empty. See [code snippet](https://github.com/getsentry/sentry/pull/77762/commits/eca1b546d0a333cd71cfd794415bf0fc084f35ab#diff-b807ec9bd1ce652df652103bcbc5b8712969950c27819ece2e4c5fdad78d434cR38). 

Is there a better practice we can take here by altering the `GetTaskResponse` schema?

proto/sentry_protos/sentry/v1alpha/taskworker.proto
```diff
@@ -0,0 +1,141 @@
...
+  optional google.protobuf.Timestamp deadline = 6;
+
+  // Retry state
+  RetryState retry_state = 7;
+}
+
+// Once a WorkerTask has been received by the task consumer it is 'upgraded'
+// into a PendingTask so that processing state can be tracked.
+// This proto might not be used as Pending tasks don't need to cross process boundaries.
+message PendingTask {
```
enochtangg, line 49: Is the reason why we don't need to store information like topic name and partition because topic namespaces are assigned to specific worker groups? Therefore, this information is implicitly known?

enochtangg (commented):
Changes look great. Just had a few questions for you. 

Also FYI: @john-z-yang and I prototyped the pull model last week and this was the first-pass proto definition we landed on: https://github.com/getsentry/sentry-protos/compare/main...worker-pull

proto/sentry_protos/sentry/v1alpha/taskworker.proto
markstory: We could use `optional` which would let us use `HasField` to check for a task.

proto/sentry_protos/sentry/v1alpha/taskworker.proto
markstory: Yes, a few of the storage options we're considering could have storage partitioned by topic/partition. I thought having that information in PendingTasks would be redundant, but if we go with a storage solution that requires topic/partition we can add those fields.

proto/sentry_protos/sentry/v1alpha/taskworker.proto
markstory: What about `UpdateTask`?

enochtangg (approved):

PR #40, opened 2024-10-03
feat(eap): Add uint64 to attribute value

Zylphrex:
Some values like project_id are uint64 under the hood and would not work with a int64.


mitsuhiko:
I wonder if it would not be more correct to carry IDs as strings and avoid this issue entirely.  I worry that quite a few SDKs are going to have troubles supporting both i64 and u64 in the same place without making a mess and this part of the protocol might leak upwards.

@markstory thoughts on this?

PR #43, opened 2024-10-04
feat(filtering): Add float array

wmak:
- Add support for float arrays
- Don't have an usecase for it, but should we support boolean arrays too?

### DIFF
--- a/proto/sentry_protos/snuba/v1/trace_item_attribute.proto
+++ b/proto/sentry_protos/snuba/v1/trace_item_attribute.proto
@@ -37,6 +37,10 @@ message IntArray {
   repeated int64 values = 1;
 }
 
+message FloatArray {
+  repeated float values = 1;
+}
+
 message AttributeValue {
   oneof value {
     bool val_bool = 1;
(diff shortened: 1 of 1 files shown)


volokluev:
What's the float array usecase?

wmak:
Numeric attributes are floats, and can be effectively used for any numeric values that users might want to segment by. 

A sentry example of this is we currently tag the duration of the query, eg. one day queries will store `86400`. Being able to query for example `query.period:[86400,604800]` would be a valid query

volokluev:
If you're tagging something, use a string. numerical attributes are for aggregations and approximations

wmak:
Are users not able to set numeric attributes?

Also aren't numeric attributes always floats? Why do we have `IntArray` but can't add `FloatArray`?

mitsuhiko:
Since we're generally following open telemetry here is the full scope of what is permissible there:

* a primitive: string, boolean, double precision floating point, signed 64 bit integer
* a homogeneous array of any primitive

So we should generally assume that all of these need supporting.

evanh (approved):

PR #44, opened 2024-10-08
feat: Publish documentation on each release

markstory:
Publish simple generated documentation on each release. The docs are a bit sparse right now, but we can use markdown in comment blocks for messages/services to improve the documentation.


volokluev (approved):
Hell yeah, thank you

PR #46, opened 2024-10-09
feat(eap): Add an endpoint to find traces

evanh:
This endpoint allows finding traces that contain events with specific attributes.

The TraceFilter and TraceItemFilter can be combined to search using multiple events,
and multiple attributes on each event.

### DIFF
--- a/proto/sentry_protos/snuba/v1/endpoint_find_traces.proto
+++ b/proto/sentry_protos/snuba/v1/endpoint_find_traces.proto
@@ -12,10 +12,24 @@ enum TraceOrderBy {
   TRACE_ORDER_BY_TRACE_DURATION = 2;
 }
 
+message TracePageToken {
+  // This class exists to avoid circular imports with the PageToken class
+  // The next version of the proto will have a PageToken class that can be used
+  // for all endpoints, and this class can be removed.
+  oneof value {
+    uint64 offset = 1;
+    // Instead of using offset (which requires all the scanning and ordering),
+    // the server sends back a filter clause to be added on to the filter conditions
+    // which skips the previous results altogether, avoiding extra scanning and sorting
+    EventFilter event_filter = 2;
+    TraceFilter trace_filter = 3;
+  }
+}
+
 message FindTracesRequest {
   RequestMeta meta = 1;
   TraceFilter filter = 2;
-  PageToken page_token = 3;
+  TracePageToken page_token = 3;
   TraceOrderBy order_by = 4;
 }
 
@@ -27,5 +41,35 @@ message TraceResponse {
 
 message FindTracesResponse {
   repeated TraceResponse traces = 1;
-  PageToken page_token = 2;
+  TracePageToken page_token = 2;
+}
+
+message EventFilter {
+  TraceItemName trace_item_name = 1;
+  TraceItemFilter filter = 2;
+}
+
+message AndTraceFilter {
+  repeated TraceFilter filters = 1;
+}
+
+message OrTraceFilter {
+  repeated TraceFilter filters = 1;
+}
+
+message NotTraceFilter {
+  repeated TraceFilter filters = 1;
+}
+
+// Represents a set of conditions for finding particular events
+// in a trace. Each EventFilter is meant to find one particular
+// type of event. Those can then be combined to find traces that
+// contain different combinations of events.
+message TraceFilter {
+  oneof filter {
+    AndTraceFilter and_filter = 1;
+    OrTraceFilter or_filter = 2;
+    NotTraceFilter not_filter = 3;
+    EventFilter event_filter = 4;
+  }
 }
--- a/proto/sentry_protos/snuba/v1/endpoint_find_traces.proto
+++ b/proto/sentry_protos/snuba/v1/endpoint_find_traces.proto
@@ -1,0 +1,45 @@
+syntax = "proto3";
+
+package sentry_protos.snuba.v1;
+
+import "sentry_protos/snuba/v1/request_common.proto";
+import "sentry_protos/snuba/v1/trace_item_filter.proto";
+
+message EventFilter {
+  string trace_item_name = 1;
+  TraceItemFilter filter = 2;
+}
+
+message AndTraceFilter {
+  repeated EventFilter filters = 1;
+}
+
+message OrTraceFilter {
+  repeated EventFilter filters = 1;
+}
+
+message NotTraceFilter {
+  repeated EventFilter filters = 1;
+}
+
+message TraceFilter {
+  oneof value {
+    AndTraceFilter and_filter = 1;
+    OrTraceFilter or_filter = 2;
+    NotTraceFilter not_filter = 3;
+    EventFilter event_filter = 4;
+  }
+}
+
+message FindTracesRequest {
+  RequestMeta meta = 1;
+  TraceFilter filter = 2;
+}
+
+message TraceResponse {
+  string trace_id = 1;
+}
+
+message FindTracesResponse {
+  repeated TraceResponse traces = 1;
+}
--- a/proto/sentry_protos/snuba/v1/endpoint_find_traces.proto
+++ b/proto/sentry_protos/snuba/v1/endpoint_find_traces.proto
@@ -6,27 +6,6 @@ import "google/protobuf/timestamp.proto";
 import "sentry_protos/snuba/v1/request_common.proto";
 import "sentry_protos/snuba/v1/trace_item_filter.proto";
 
-message AndTraceFilter {
-  repeated TraceFilter filters = 1;
-}
-
-message OrTraceFilter {
-  repeated TraceFilter filters = 1;
-}
-
-message NotTraceFilter {
-  repeated TraceFilter filters = 1;
-}
-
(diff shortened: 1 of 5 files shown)


proto/sentry_protos/snuba/v1/endpoint_find_traces.proto
```diff
@@ -0,0 +1,45 @@
...
+    EventFilter event_filter = 4;
+  }
+}
+
+message FindTracesRequest {
+  RequestMeta meta = 1;
+  TraceFilter filter = 2;
+}
+
+message TraceResponse {
```
volokluev, line 39: you probably want to add a page token here similar to the trace item requests

proto/sentry_protos/snuba/v1/endpoint_find_traces.proto
```diff
@@ -0,0 +1,45 @@
...
+  repeated EventFilter filters = 1;
+}
+
+message OrTraceFilter {
+  repeated EventFilter filters = 1;
+}
+
+message NotTraceFilter {
+  repeated EventFilter filters = 1;
+}
```
volokluev, line 23: these make it so you can't OR/AND anything except EventFilters. You probably want 

```
message AndTraceFilter {
  repeated TraceFilter filters = 1;
}
```

Otherwise you will only be able to express oneof (AndTraceFilter, OrTraceFilter, NotTraceFilter, EventFilter)

evanh: So I actually collapsed Trace/EventFilter, I think that fixes this though.

volokluev: I don't think I understand, how would you be able to express conditions like

`(A OR B) AND (C OR D)`

With this structure? Or is that purposefully left out?

proto/sentry_protos/snuba/v1/examples.py
```diff
@@ -263,3 +272,141 @@ def test_example_table_with_aggregations() -> None:
...
+            ),
+        ),
+    )
+
+    # Find traces that contain at least one of: a span with a `span_name` of
+    # "database_query" and an error with a `group_id` of "1123"
+    FindTracesRequest(
+        meta=COMMON_META,
+        filter=TraceFilter(
+            and_filter=OrTraceFilter(
```
volokluev, line 374: did you mean
```suggestion
            and_filter=AndTraceFilter(
```

proto/sentry_protos/snuba/v1/endpoint_find_traces.proto
evanh:
```
AndTraceFilter(
    OrTraceFilter(
        TraceFilter(A), 
        TraceFilter(B),
    ),
    OrTraceFilter(
        TraceFilter(C),
        TraceFilter(D),
    ),
)
```

evanh: I think I have this fixed now.

proto/sentry_protos/snuba/v1/endpoint_find_traces.proto
```diff
@@ -0,0 +1,50 @@
...
+message FindTracesRequest {
+  RequestMeta meta = 1;
+  TraceFilter filter = 2;
+  PageToken page_token = 3;
+}
+
+message TraceResponse {
+  string trace_id = 1;
+  google.protobuf.Timestamp start_timestamp = 5;
+  google.protobuf.Timestamp end_timestamp = 6;
```
volokluev, line 44: what are these for?

evanh: Those denote the time range of just that trace. So for each trace the response has the ID of the trace, and the start/end timestamp for that trace.

proto/sentry_protos/snuba/v1/endpoint_find_traces.proto
```diff
@@ -0,0 +1,50 @@
...
+    OrTraceFilter or_filter = 2;
+    NotTraceFilter not_filter = 3;
+    EventFilter event_filter = 4;
+  }
+}
+
+message FindTracesRequest {
+  RequestMeta meta = 1;
+  TraceFilter filter = 2;
+  PageToken page_token = 3;
```
volokluev, line 38: I don't think this page token object is going to quite work. Mostly because it only consists of `TraceItemFilter`s or `offset`s. Which is not sufficient for paginating through traces.

evanh: That's an interesting point:

In order to paginate the results, the filter could be applied either to the result set of traces or pushed down to the event tables.

In the first case, the filter would be something like `X < trace_id < Y` or `X < end_timestamp < Y`. I don't know if either is a good idea, since it will require building the entire result set in order to filter.

In the second case, the filter would be pushed down to the event tables, which means a `TraceItemFilter` actually does make sense.

volokluev: What you are saying makes sense. Maybe the `PageToken` needs to have `EventFilter` instead of `TraceItemFilter` as its filter attribute. Because as of now since `TraceItemFilter` is not tied to the actual `trace_item_name`

evanh: I added this to the PageToken, which also required reorganizing the files a bit.

volokluev (approved):
LGTM, performance team should review as well though as they are the experts in finding traces

proto/sentry_protos/snuba/v1/examples.py
```diff
@@ -263,3 +272,165 @@ def test_example_table_with_aggregations() -> None:
...
+                            ),
+                        ]
+                    ),
+                ),
+            ),
+        ),
+    )
+
+    # Find traces that contain two events: a span with a `span_name` of
+    # "database_query" and an error with a `group_id` of "1123"
```
Zylphrex, line 332: Having 2 trace filters for 2 different spans in the same trace is a valid use case as well right?

proto/sentry_protos/snuba/v1/trace_item_filter.proto
```diff
@@ -50,3 +50,35 @@ message TraceItemFilter {
     ExistsFilter exists_filter = 5;
   }
 }
+
+// Wraps a TraceItemFilter and specifies the trace item name,
+// to be used for cross-event queries.
+message EventFilter {
+  string trace_item_name = 1;
```
Zylphrex, line 57: What are all the possible options for `trace_item_name`? From the examples, I see we have `spans` and `errors`. Are there more? Can this be an enum?

proto/sentry_protos/snuba/v1/endpoint_find_traces.proto
```diff
@@ -0,0 +1,24 @@
+syntax = "proto3";
+
+package sentry_protos.snuba.v1;
+
+import "google/protobuf/timestamp.proto";
+import "sentry_protos/snuba/v1/request_common.proto";
+import "sentry_protos/snuba/v1/trace_item_filter.proto";
+
+message FindTracesRequest {
```
Zylphrex, line 9: What will the sort order of the response be? The ask is to have it sortable by timestamp or trace duration, with timestamp being the more common sort.

proto/sentry_protos/snuba/v1/examples.py
evanh: Yep!

proto/sentry_protos/snuba/v1/trace_item_filter.proto
evanh: Ultimately it would be one for each event type that can be connected to a trace (replays, profiles, functions).

Having an Enum makes sense, we can change that.

@volokluev Can I change that in v1 or should I wait until v2?

proto/sentry_protos/snuba/v1/endpoint_find_traces.proto
evanh: I'll add something to change the sort order. This API spec makes no distinction on how the results are sorted, so we can do that when we actually implement the endpoint.

proto/sentry_protos/snuba/v1/trace_item_filter.proto
volokluev: There already is an enum:

https://github.com/getsentry/sentry-protos/blob/main/proto/sentry_protos/snuba/v1/request_common.proto#L19

Zylphrex (approved):

proto/sentry_protos/snuba/v1/endpoint_find_traces.proto
```diff
@@ -0,0 +1,31 @@
...
+}
+
+message FindTracesRequest {
+  RequestMeta meta = 1;
+  TraceFilter filter = 2;
+  PageToken page_token = 3;
+  TraceOrderBy order_by = 4;
+}
+
+message TraceResponse {
```
Zylphrex, line 22: One thing I forgot to mention is we should talk about what's possible to return here as this will be used to render the trace samples table. But those are additional attributes which should be backwards compatible with the changes here.

PR #51, opened 2024-10-19
Add QueryStats and QueryMetadata to TraceItemTableResponse

nachivrn:
Add QueryStats and QueryMetadata to API endpoints response

    - Update proto/sentry_protos/snuba/v1/endpoint_trace_item_table.proto
      to include optional QueryStats and QueryMetadata fields in
      TraceItemTableResponse
    - Modify proto/sentry_protos/snuba/v1/request_common.proto
      to define QueryStats and QueryMetadata message structures

    These changes allow for more detailed query execution information
    to be included in responses, enhancing observability and debugging capabilities.

### DIFF
--- a/proto/sentry_protos/snuba/v1/endpoint_trace_item_table.proto
+++ b/proto/sentry_protos/snuba/v1/endpoint_trace_item_table.proto
@@ -21,7 +21,6 @@ message TraceItemTableRequest {
   uint32 limit = 6;
   PageToken page_token = 7;
   repeated VirtualColumnContext virtual_column_contexts = 8;
-  bool debug = 10;
 }
 
 message Column {
@@ -40,6 +39,5 @@ message TraceItemColumnValues {
 message TraceItemTableResponse {
   repeated TraceItemColumnValues column_values = 1;
   PageToken page_token = 2;
-  optional QueryStats query_stats = 3;
-  optional QueryMetadata query_metadata = 4;
+  ResponseMeta meta = 3;
 }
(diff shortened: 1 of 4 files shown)


proto/sentry_protos/snuba/v1/endpoint_trace_item_table.proto
```diff
@@ -21,6 +21,7 @@ message TraceItemTableRequest {
   uint32 limit = 6;
   PageToken page_token = 7;
   repeated VirtualColumnContext virtual_column_contexts = 8;
+  bool debug = 10;
```
nachivrn, line 24: Debug flag will be used to add QueryStats and QueryMetadata in the response

proto/sentry_protos/snuba/v1/endpoint_trace_item_table.proto
```diff
@@ -39,4 +40,6 @@ message TraceItemColumnValues {
 message TraceItemTableResponse {
   repeated TraceItemColumnValues column_values = 1;
   PageToken page_token = 2;
+  optional QueryStats query_stats = 3;
+  optional QueryMetadata query_metadata = 4;
```
evanh, line 44: I would put both of these objects inside another object, so that you only have to add one field to the response instead of two.

proto/sentry_protos/snuba/v1/endpoint_trace_item_table.proto
evanh: This should go in the `RequestMeta` object so it can be generally available.

evanh (commented):
I wonder if it's better for our next version of the API to have a `ResponseMeta`, similar to the `RequestMeta` that every endpoint returns. That would make it easier to have these common return objects in every endpoint.

proto/sentry_protos/snuba/v1/endpoint_trace_item_table.proto
volokluev: these should also be repeated because there could be many queries in one request

volokluev (approved):

evanh (approved):

PR #54, opened 2024-10-25
Add trace logging support to ResponseMeta

nachivrn:
This commit introduces trace logging functionality:

- Added `trace_log` string field to ResponseMeta

These changes allow clients to request trace logging for their queries and receive the clickhouse trace logs in the response. This feature will help with debugging and performance analysis of EAP API.

### DIFF
--- a/proto/sentry_protos/snuba/v1/request_common.proto
+++ b/proto/sentry_protos/snuba/v1/request_common.proto
@@ -16,6 +16,7 @@ message RequestMeta {
   TraceItemName trace_item_name = 7;
   bool debug = 10;
   string request_id = 11;
+  bool enable_trace_logs = 12;
 }
 
 message ResponseMeta {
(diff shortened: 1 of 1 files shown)


proto/sentry_protos/snuba/v1/request_common.proto
```diff
@@ -16,6 +16,7 @@ message RequestMeta {
   TraceItemName trace_item_name = 7;
   bool debug = 10;
   string request_id = 11;
+  bool enable_trace_logs = 12;
```
volokluev, line 19: why is this not just the `debug` flag?

nachivrn: Initially, I considered using enable_trace_logs to control both trace output capture and profile event logging. I have updated to use a single flag.

kylemumma (approved):

kylemumma:
`trace_logs` refers to clickhouse trace logs

PR #56, opened 2024-10-28
feat(taskworker) Use uint instead of int

markstory:
Retry state and offsets won't have negative numbers.


enochtangg (approved):

john-z-yang (approved):

markstory:
Closing as we have more changes to make to these structs and will need to change the version/module name.

PR #60, opened 2024-11-04
feat(subscriptions): Create subscriptions uses timeseries request

shruthilayaj:
We decided to use time series instead of trace item table request in subscriptions.
This protobuf isn't used anywhere right now.


proto/sentry_protos/snuba/v1/endpoint_create_subscription.proto
```diff
@@ -0,0 +1,16 @@
+syntax = "proto3";
+
+package sentry_protos.snuba.v1;
+
+import "sentry_protos/snuba/v1/endpoint_time_series.proto";
+
+message CreateSubscriptionRequest {
+  TimeSeriesRequest time_series_request = 1;
+  uint64 project_id = 2;
```
volokluev, line 9: The TimeSeriesRequest already has this, you don't need to put it here

edwardgou-sentry (approved):

PR #61, opened 2024-11-05
feat(tasks): Add protos for the ConsumerService

evanh:
Add proto definitions for the ConsumerService as well as protos for the messages used
to interact with that service.

Add a common error structure for returning errors.

The service currently has support for GetTask and SetTaskStatus, which is what the workers
will use to fetch new tasks and update the task statuses after a task is finished.

### DIFF
--- a/proto/sentry_protos/sentry/v1/taskworker.proto
+++ b/proto/sentry_protos/sentry/v1/taskworker.proto
@@ -109,8 +109,7 @@ service ConsumerService {
   // Fetch a new task activation to process.
   rpc GetTask(GetTaskRequest) returns (GetTaskResponse) {}
 
-  // Update the state of a task with execution results. (also used in push
-  // models)
+  // Update the state of a task with execution results.
   rpc SetTaskStatus(SetTaskStatusRequest) returns (SetTaskStatusResponse) {}
 }
 
@@ -120,7 +119,6 @@ message GetTaskRequest {
 message GetTaskResponse {
   // If there are no tasks available, these will be empty
   optional TaskActivation task = 1;
-  optional google.protobuf.Timestamp processing_deadline = 2;
 
   optional Error error = 3;
 }
@@ -136,8 +134,6 @@ message SetTaskStatusRequest {
 message SetTaskStatusResponse {
   // The next task the worker should execute. Requires fetch_next=True on the request.
   optional TaskActivation task = 1;
-  // The processing deadline for the new task
-  optional google.protobuf.Timestamp processing_deadline = 2;
 
   optional Error error = 3;
 }
--- a/proto/sentry_protos/sentry/v1/taskworker.proto
+++ b/proto/sentry_protos/sentry/v1/taskworker.proto
@@ -2,6 +2,7 @@ syntax = "proto3";
 
 package sentry_protos.sentry.v1;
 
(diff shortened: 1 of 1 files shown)


proto/sentry_protos/sentry/v1/taskworker.proto
```diff
@@ -91,3 +92,52 @@ message InflightActivation {
...
+  rpc SetTaskStatus(SetTaskStatusRequest) returns (SetTaskStatusResponse) {}
+}
+
+message GetTaskRequest {
+  optional string namespace = 1;
+}
+message GetTaskResponse {
+  // If there are no tasks available, these will be empty
+  optional TaskActivation task = 1;
+  optional google.protobuf.Timestamp processing_deadline = 2;
```
markstory, line 123: We might not need this now. `TaskActivation` includes `processing_deadline_duration` now which will let workers figure out the deadline once they get the response of `GetTask`.

proto/sentry_protos/sentry/v1/taskworker.proto
```diff
@@ -91,3 +92,52 @@ message InflightActivation {
...
+
+  optional Error error = 3;
+}
+
+message SetTaskStatusRequest {
+  string id = 1;
+  string namespace = 2;
```
markstory, line 130: Do we need `namespace` if we have the `id` of the activation?

proto/sentry_protos/sentry/v1/taskworker.proto
```diff
@@ -91,3 +92,52 @@ message InflightActivation {
...
+  TaskActivationStatus status = 3;
+
+  // Set to true to have receive a new task in the response
+  optional bool fetch_next = 4;
+}
+message SetTaskStatusResponse {
+  // The next task the worker should execute. Requires fetch_next=True on the request.
+  optional TaskActivation task = 1;
+  // The processing deadline for the new task
+  optional google.protobuf.Timestamp processing_deadline = 2;
```
markstory, line 140: I'm not sure we'll need this as `TaskActivation` has `processing_deadline_duration` built in now. When we added this for the prototypes, `TaskActivation` did not have a processing duration builtin.

proto/sentry_protos/sentry/v1/taskworker.proto
```diff
@@ -91,3 +92,52 @@ message InflightActivation {
...
+  // A list of messages that carry any error details.
+  repeated google.protobuf.Any details = 3;
+}
+
+service ConsumerService {
+  // Fetch a new task activation to process.
+  rpc GetTask(GetTaskRequest) returns (GetTaskResponse) {}
+
+  // Update the state of a task with execution results. (also used in push
+  // models)
```
markstory, line 113:
```suggestion
  // Update the state of a task with execution results.
```

No push model anymore.

proto/sentry_protos/sentry/v1/taskworker.proto
evanh: I included namespace here because I assumed workers would need to know which namespace they are updating, in order to make sure the write gets written back to the same consumer.

evanh: Yeah OK I'll remove this. I think I just don't understand the load balancing model well enough yet.

Issue #42, opened 2024-10-04
Provide tooling for executing examples in CI


markstory:
The protobuf -> python conversion (and rust too) is not always immediately obvious. Having example usage of message structs could help serve as a reference to improve onboarding and ease of use.  We should also execute all of the examples to ensure there are no runtime errors. `pytest` could execute example modules and validate that all the examples are valid constructions.

https://github.com/getsentry/sentry-protos/blob/main/proto/sentry_protos/snuba/v1/examples.py

is an examples file that could be made executable.

markstory:
Done now.

PR #62, opened 2024-11-06
feat: Run examples as tests and add tests for taskworker

markstory:
- Move examples into py/tests so that we can run them.
- Fix broken examples in snuba module.
- Add tests for taskworker protos
- Extend CI configuration to run python tests on generated code.

Refs #42


volokluev (approved):
king shit

PR #64, opened 2024-11-13
Add reliability to timeseries results

davidtsuk:
Average sampling rate and reliability are calculated on a per time bucket basis.

### DIFF
--- a/proto/sentry_protos/snuba/v1/endpoint_time_series.proto
+++ b/proto/sentry_protos/snuba/v1/endpoint_time_series.proto
@@ -34,10 +34,10 @@ message TimeSeries {
   repeated DataPoint data_points = 4;
   // number of events used to calculate this timeseries
   uint64 num_events = 5;
+  float avg_sampling_rate = 6 [deprecated = true];
 }
 
 message TimeSeriesResponse {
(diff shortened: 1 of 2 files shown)


proto/sentry_protos/snuba/v1/endpoint_time_series.proto
```diff
@@ -19,6 +19,8 @@ message DataPoint {
   float data = 1;
   // optional, if not set, assume true. Accounts for sparse time series
   bool data_present = 2;
+  float avg_sampling_rate = 3;
+  bool is_reliable = 4;
```
volokluev, line 23: add a comment explaining what this field is

volokluev (approved):
document the field, otherwise good

proto/sentry_protos/snuba/v1/endpoint_time_series.proto
```diff
@@ -31,7 +34,6 @@ message TimeSeries {
   repeated DataPoint data_points = 4;
   // number of events used to calculate this timeseries
   uint64 num_events = 5;
-  float avg_sampling_rate = 6;
```
volokluev, line 34: don't delete this field, deprecate it

PR #68, opened 2024-12-05
feat(taskbroker): Add namespace parameter to get/set requests

enochtangg:
In order for taskbroker to return tasks specific to namespace (e.g. symbolicator use-case), it needs to be able to receive a namespace as part of the gRPC request object.


proto/sentry_protos/sentry/v1/taskworker.proto
```diff
@@ -130,6 +132,7 @@ message SetTaskStatusRequest {
 
   // Set to true to have receive a new task in the response
   optional bool fetch_next = 4;
+  optional string fetch_next_namespace = 5;
```
markstory, line 135: Would be nice if the grpc server could either respond with a next task if this or `fetch_next` is used.

markstory (approved):

PR #69, opened 2024-12-06
feat(taskbroker): Improve fetch_next

enochtangg:
Enrich `fetch_next` type. By doing this, we would not run into the case where`fetch_next: bool` is not provided (or set to false) and `fetch_next_namespace: string` is provided. Now, either the fetch_next message is provided with an optional namespace, or not at all.


john-z-yang (approved):
I think the field numbers might need to change, but I'm happy with this

PR #71, opened 2024-12-13
feat(eap): Add a GetTraces endpoint

phacops:
This will add Protobuf files describing a new endpoint to find traces. It will return a list of traces with fields requested.

Features:
- You can pick which fields from the `enum`
- You can specify the type you want from a field
- You can use the same filter conditions as for the `TraceItemTable` endpoint
- You can order by any field from the `enum`

Limitations:
- No aggregation possible
- No support for alias
- A fixed list of fields to query

Since traces are not a defined object but a specific view on an aggregated list of spans, I don't think we should plan to support any field possible. Maybe at some point, we'll make the trace a real object and we'll add trace attributes and by then, we can design a new version of the endpoint.

Closes https://github.com/getsentry/eap-planning/issues/121.

### DIFF
--- a/proto/sentry_protos/snuba/v1/endpoint_get_traces.proto
+++ b/proto/sentry_protos/snuba/v1/endpoint_get_traces.proto
@@ -34,7 +34,6 @@ message TraceAttribute {
   }
 
   Key key = 1;
-  AttributeKey.Type type = 2;
   AttributeValue value = 3;
 }
 
--- a/proto/sentry_protos/snuba/v1/endpoint_get_traces.proto
+++ b/proto/sentry_protos/snuba/v1/endpoint_get_traces.proto
@@ -6,24 +6,25 @@ import "sentry_protos/snuba/v1/request_common.proto";
 import "sentry_protos/snuba/v1/trace_item_attribute.proto";
 import "sentry_protos/snuba/v1/trace_item_filter.proto";
 
-message TraceColumn {
-  enum Name {
-    NAME_UNSPECIFIED = 0;
-    NAME_TRACE_ID = 1;
-    NAME_START_TIMESTAMP = 2;
-    NAME_ROOT_SPAN_NAME = 3;
-    NAME_TOTAL_SPAN_COUNT = 4;
-    NAME_FILTERED_SPAN_COUNT = 5;
-    NAME_ROOT_SPAN_DURATION_MS = 6;
-  }
+enum TraceAttributeKey {
+  TRACE_ATTRIBUTE_KEY_UNSPECIFIED = 0;
+  TRACE_ATTRIBUTE_KEY_TRACE_ID = 1;
+  TRACE_ATTRIBUTE_KEY_START_TIMESTAMP = 2;
+  TRACE_ATTRIBUTE_KEY_ROOT_SPAN_NAME = 3;
+  TRACE_ATTRIBUTE_KEY_TOTAL_SPAN_COUNT = 4;
+  TRACE_ATTRIBUTE_KEY_FILTERED_SPAN_COUNT = 5;
+  TRACE_ATTRIBUTE_KEY_ROOT_SPAN_DURATION_MS = 6;
+}
 
-  Name name = 1;
+message TraceAttribute {
+  TraceAttributeKey key = 1;
   AttributeKey.Type type = 2;
+  AttributeValue value = 3;
 }
 
 message GetTracesRequest {
   message OrderBy {
-    TraceColumn column = 1;
+    TraceAttributeKey key = 1;
     bool descending = 2;
   }
 
@@ -36,22 +37,17 @@ message GetTracesRequest {
   PageToken page_token = 2;
   uint32 limit = 3;
 
-  repeated TraceItemFilter filters = 4;
+  repeated TraceFilter filters = 4;
   repeated OrderBy order_by = 5;
-  repeated TraceColumn columns = 6;
+  repeated TraceAttribute attributes = 6;
 }
 
 message GetTracesResponse {
   message Trace {
-    message Column {
-      TraceColumn.Name name = 1;
-      AttributeValue value = 2;
-    }
-
-    repeated Column columns = 1;
+    repeated TraceAttribute attributes = 1;
   }
 
-  repeated Trace traces = 1;
+  repeated Trace rows = 1;
 
   PageToken page_token = 2;
   ResponseMeta meta = 3;
--- a/proto/sentry_protos/snuba/v1/endpoint_get_traces.proto
+++ b/proto/sentry_protos/snuba/v1/endpoint_get_traces.proto
@@ -34,7 +34,11 @@ message TraceAttribute {
   }
 
   Key key = 1;
-  AttributeValue value = 3;
+  AttributeValue value = 2;
+
+  // AttributeKey.Type will specify the type of the attribute we return.
+  // It does not need to be sent when requesting an attribute.
+  AttributeKey.Type type = 3;
 }
 
 // GetTracesRequest lets you query traces with various attributes.
--- a/proto/sentry_protos/snuba/v1/endpoint_get_traces.proto
+++ b/proto/sentry_protos/snuba/v1/endpoint_get_traces.proto
@@ -28,6 +28,9 @@ message TraceAttribute {
     // KEY_ROOT_SPAN_DURATION_MS will return the duration of the root span in
     // milliseconds.
     KEY_ROOT_SPAN_DURATION_MS = 6;
+    // KEY_ROOT_SPAN_PROJECT_ID returns the project ID associated with the root
+    // span.
+    KEY_ROOT_SPAN_PROJECT_ID = 7;
   }
 
   Key key = 1;
--- a/proto/sentry_protos/snuba/v1/endpoint_get_traces.proto
+++ b/proto/sentry_protos/snuba/v1/endpoint_get_traces.proto
@@ -1,0 +1,52 @@
+syntax = "proto3";
+
+package sentry_protos.snuba.v1;
+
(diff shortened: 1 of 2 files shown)


proto/sentry_protos/snuba/v1/endpoint_get_traces.proto
```diff
@@ -0,0 +1,52 @@
+syntax = "proto3";
+
+package sentry_protos.snuba.v1;
+
+import "sentry_protos/snuba/v1/request_common.proto";
+import "sentry_protos/snuba/v1/trace_item_attribute.proto";
+import "sentry_protos/snuba/v1/trace_item_filter.proto";
+
+message TraceColumn {
```
colin-sentry, line 9: why not just use attributekey here?

proto/sentry_protos/snuba/v1/endpoint_get_traces.proto
```diff
@@ -0,0 +1,52 @@
...
+  TraceItemFilter filter = 2;
+  PageToken page_token = 3;
+  uint32 limit = 4;
+
+  repeated OrderBy order_by = 5;
+  repeated TraceColumn columns = 6;
+}
+
+message GetTracesResponse {
+  message Trace {
```
colin-sentry, line 39: this type is also unclear to me - why does a trace have columns?

colin-sentry: and why doesn't it have a UUID?

phacops: It's just a mechanism to only get back what you asked for. I defined a `enum` with the fields you can request from this API and the `columns` field is a list of the values you requested.

One of the value you can request is `GetTraces.Column.NAME_TRACE_ID` which will return a UUID of the `trace_id`.

proto/sentry_protos/snuba/v1/endpoint_get_traces.proto
phacops: I wanted something more specific than `AttributeKey` since we don't have dynamic attributes support and we want some arbitrary fields like `ROOT_SPAN_NAME` or the different span counts.

We can't aggregate traces arbitrarily like we do for spans, so having a specific list of what we can support seemed more useful than making it completely dynamic.

proto/sentry_protos/snuba/v1/endpoint_get_traces.proto
```diff
@@ -0,0 +1,52 @@
...
+import "sentry_protos/snuba/v1/trace_item_filter.proto";
+
+message TraceColumn {
+  enum Name {
+    NAME_UNSPECIFIED = 0;
+    NAME_TRACE_ID = 1;
+    NAME_START_TIMESTAMP = 2;
+    NAME_ROOT_SPAN_NAME = 3;
+    NAME_TOTAL_SPAN_COUNT = 4;
+    NAME_FILTERED_SPAN_COUNT = 5;
```
davidtsuk, line 16: what is filtered span count?

proto/sentry_protos/snuba/v1/endpoint_get_traces.proto
```diff
@@ -0,0 +1,52 @@
+syntax = "proto3";
+
+package sentry_protos.snuba.v1;
+
+import "sentry_protos/snuba/v1/request_common.proto";
+import "sentry_protos/snuba/v1/trace_item_attribute.proto";
+import "sentry_protos/snuba/v1/trace_item_filter.proto";
+
+message TraceColumn {
+  enum Name {
```
colin-sentry, line 10: just a few things:
- rename some of these, I would move `enum Name` to top-level scope and call it something like `enum TraceMetadataType` or `TraceKey` or `TraceSelect`
- (subjective) rename Trace in response to just be Row
- (subjective) the whole request might be better called 'TraceTableRequest' or something like that
- add docstrings to at least the top-level messages, and each enum member

proto/sentry_protos/snuba/v1/endpoint_get_traces.proto
phacops: `FILTERED_SPAN_COUNT` is the number of spans respecting the conditions where `TOTAL_SPAN_COUNT` is always the count of spans in a trace, regardless of conditions.

kylemumma (approved):
I've been going back through our proto definitions and adding better documentation on each of the `Request` and `Response` protos, so users can understand what the endpoints are and how to use them. I would appreciate if you could add these comments for documentation
1. Add a comment to the request object that describes the endpoint
2. add comments to each of the inputs in the request object to describe what it is (if non obvious)
3. add comments to each field of the response object to describe what it is (if non obvious)

proto/sentry_protos/snuba/v1/endpoint_get_traces.proto
```diff
@@ -0,0 +1,52 @@
...
+}
+
+message GetTracesRequest {
+  message OrderBy {
+    TraceColumn column = 1;
+    bool descending = 2;
+  }
+
+  RequestMeta meta = 1;
+  TraceItemFilter filter = 2;
```
volokluev, line 30: this won't work for searching across event types. At least not by itself. The filter currently does not support a `name`. That's why in the other file there's extra filter definitions for traces (because they are different).

phacops: I didn't want to enable filtering on traces, I want to only allow filtering on trace items, that's why I removed it.

Traces don't have a name, they don't have attributes, I don't think we should enable filtering on things that are defined by trace item fields.

volokluev: with your current API, how would you find all traces where `span.op = http` and `error.project_id=1`?

phacops: Pass a `TraceItemFilter` with your conditions, we map the prefixes to each dataset and we query.

volokluev: or event better, show me an example with python

phacops: I think this should fix your concerns: https://github.com/getsentry/sentry-protos/pull/71/commits/ce87464fd6b10afabc9da96031963c2ea52ffdff .

proto/sentry_protos/snuba/v1/endpoint_get_traces.proto
phacops: Why move the enum to top-level? It's really meant to be used only in this endpoint for a request, why not keep it embedded?

proto/sentry_protos/snuba/v1/endpoint_get_traces.proto
```diff
@@ -0,0 +1,76 @@
...
+message TraceAttribute {
+  // Key lists the available trace attribute keys you can query.
+  enum Key {
+    KEY_UNSPECIFIED = 0;
+    KEY_TRACE_ID = 1;
+    // KEY_START_TIMESTAMP will return the earliest timestamp seen in the trace.
+    KEY_START_TIMESTAMP = 2;
+    // KEY_ROOT_SPAN_NAME will return the name of the root (segment) span of
+    // the trace.
+    KEY_ROOT_SPAN_NAME = 3;
```
Zylphrex, line 19: Will this have the heuristics we spoke about built in or will there be other attributes for those cases?

proto/sentry_protos/snuba/v1/endpoint_get_traces.proto
```diff
@@ -0,0 +1,76 @@
...
+  RequestMeta meta = 1;
+  PageToken page_token = 2;
+  uint32 limit = 3;
+
+  // List of filters on items of the trace we'll use when querying.
+  repeated TraceFilter filters = 4;
+  // List of attributes we'd like to order by.
+  repeated OrderBy order_by = 5;
+  // List of attributes we want to query.
+  repeated TraceAttribute attributes = 6;
```
Zylphrex, line 61: Does this mean we need to specify the type of every attribute when querying? A little inconvenient but not a deal breaker.

proto/sentry_protos/snuba/v1/endpoint_get_traces.proto
```diff
@@ -0,0 +1,76 @@
...
+package sentry_protos.snuba.v1;
+
+import "sentry_protos/snuba/v1/request_common.proto";
+import "sentry_protos/snuba/v1/trace_item_attribute.proto";
+import "sentry_protos/snuba/v1/trace_item_filter.proto";
+
+// TraceAttribute represents the attribute of a trace.
+message TraceAttribute {
+  // Key lists the available trace attribute keys you can query.
+  enum Key {
```
Zylphrex, line 12: We'll also want a `KEY_ROOT_SPAN_PROJECT`. Just the project id is fine

proto/sentry_protos/snuba/v1/endpoint_get_traces.proto
phacops: This won't have the heuristic of finding either the root span name or the first http span name. I think we should add another attribute key for that and you can choose which one you want.

proto/sentry_protos/snuba/v1/endpoint_get_traces.proto
Zylphrex: We'll likely want it in the response? So I'm okay with keeping it for now

Zylphrex (approved):

davidtsuk (approved):

proto/sentry_protos/snuba/v1/endpoint_get_traces.proto
phacops: I added it back.

PR #81, opened 2025-01-06
feat(eap): Add aggregation filter

davidtsuk:
Adding an aggregation filter to allow queries with a having clause in order to support https://github.com/getsentry/eap-planning/issues/143

# Example
A trace item table request with an aggregation filter could look something like this:
```
{
  "meta": {
  },
  "columns": [
    {
      "key": {
        "type": "TYPE_STRING",
        "name": "sentry.browser.name"
      },
      "label": "browser.name"
    },
    {
      "aggregation": {
        "aggregate": "FUNCTION_AVG",
        "key": {
          "type": "TYPE_FLOAT",
          "name": "sentry.duration"
        },
        "label": "avg(duration)",
        "extrapolation_mode": "EXTRAPOLATION_MODE_SAMPLE_WEIGHTED"
      },
      "label": "avg(duration)"
    }
  ],
  "order_by": [
    {
      "column": {
        "aggregation": {
          "aggregate": "FUNCTION_AVG",
          "key": {
            "type": "TYPE_FLOAT",
            "name": "sentry.duration"
          },
          "label": "avg(duration)",
          "extrapolation_mode": "EXTRAPOLATION_MODE_SAMPLE_WEIGHTED"
        }
      }
    }
  ],
  "aggregation_filter": {
    "comparison_filter": {
      "aggregate": "FUNCTION_COUNT",
      "key": {
        "type": "TYPE_FLOAT",
        "name": "sentry.duration"
      },
      "op": "OP_GREATER_THAN",
      "value": {
        "val_int": 100
      }
    }
  },
  "limit": 100
}
```

### DIFF
--- a/proto/sentry_protos/snuba/v1/endpoint_trace_item_table.proto
+++ b/proto/sentry_protos/snuba/v1/endpoint_trace_item_table.proto
@@ -36,6 +36,7 @@ message TraceItemTableRequest {
 
   repeated VirtualColumnContext virtual_column_contexts = 8;
 
+  // filter out results using aggregate functions
   AggregationFilter aggregation_filter = 9;
 }
 
--- a/proto/sentry_protos/snuba/v1/endpoint_trace_item_table.proto
+++ b/proto/sentry_protos/snuba/v1/endpoint_trace_item_table.proto
@@ -35,6 +35,40 @@ message TraceItemTableRequest {
   PageToken page_token = 7;
 
   repeated VirtualColumnContext virtual_column_contexts = 8;
+
+  AggregationFilter aggregation_filter = 9;
+}
+
+message AggregationAndFilter {
+  repeated AggregationFilter filters = 1;
+}
+
+message AggregationOrFilter {
+  repeated AggregationFilter filters = 1;
+}
+
+message AggregationComparisonFilter {
+  enum Op {
+    OP_UNSPECIFIED = 0;
+    OP_LESS_THAN = 1;
+    OP_GREATER_THAN = 2;
+    OP_LESS_THAN_OR_EQUALS = 3;
+    OP_GREATER_THAN_OR_EQUALS = 4;
+    OP_EQUALS = 5;
+    OP_NOT_EQUALS = 6;
+  }
+  Function aggregate = 1;
+  AttributeKey key = 2;
+  Op op = 3;
+  AttributeValue value = 4;
+}
+
+message AggregationFilter {
+  oneof value {
+    AggregationAndFilter and_filter = 1;
+    AggregationOrFilter or_filter = 2;
+    AggregationComparisonFilter comparison_filter = 3;
+  }
 }
 
 message Column {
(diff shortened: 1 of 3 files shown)


volokluev:
1. When making protocol changes, please add an example to make sure you've done your due dilligence with how the interface will actually function
2. I don't think putting this in the traceitemfilter is the right way to go. AFAICS, this is only applicable to `TraceItemTable`, it does not make sense for `TimeSeries` queries or `FindTraces` queries. In clickhouse this is a `HAVING` clause. it is separate because it is a post-aggregation step, I don't think we can get away from that reality here

davidtsuk:
> 1. When making protocol changes, please add an example to make sure you've done your due dilligence with how the interface will actually function
Will do!
> 2. I don't think putting this in the traceitemfilter is the right way to go. AFAICS, this is only applicable to `TraceItemTable`, it does not make sense for `TimeSeries` queries or `FindTraces` queries. In clickhouse this is a `HAVING` clause. it is separate because it is a post-aggregation step, I don't think we can get away from that reality here
That's a good point, we probably don't need to add this feature to timeseries queries. I'll move it to `endpoint_trace_item_table`

proto/sentry_protos/snuba/v1/endpoint_trace_item_table.proto
```diff
@@ -35,6 +35,41 @@ message TraceItemTableRequest {
   PageToken page_token = 7;
 
   repeated VirtualColumnContext virtual_column_contexts = 8;
+
+  // filter out results using aggregate functions
```
volokluev, line 39: I miscommunicated about the examples. What I really meant is that you should write tests:

https://github.com/getsentry/sentry-protos/tree/main/py/tests

We already have a decent amount for the protocols. They act as automatically enforced documentation

proto/sentry_protos/snuba/v1/endpoint_trace_item_table.proto
```diff
@@ -35,6 +35,41 @@ message TraceItemTableRequest {
...
+    OP_UNSPECIFIED = 0;
+    OP_LESS_THAN = 1;
+    OP_GREATER_THAN = 2;
+    OP_LESS_THAN_OR_EQUALS = 3;
+    OP_GREATER_THAN_OR_EQUALS = 4;
+    OP_EQUALS = 5;
+    OP_NOT_EQUALS = 6;
+  }
+  Function aggregate = 1;
+  AttributeKey key = 2;
```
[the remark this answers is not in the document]
davidtsuk: Is there a case where someone would want to apply a post-aggregation filter that compares against a regular column? From a performance perspective, I don't think it ever makes sense to use a post-aggregation filter instead of a pre-aggregation filter with individual columns/attributes. I agree that it would be cleaner to use a Column, but I also think that having a more restrictive API would prevent users from unintentionally crafting slower queries.

Good point about the extrapolated and non-extrapolated aggregations, the extrapolation mode should also be specified for the post aggregation filter.

proto/sentry_protos/snuba/v1/endpoint_trace_item_table.proto
davidtsuk: Ah, I wasn't aware that we had tests in this repo, thanks for the clarification

volokluev (approved):

PR #83, opened 2025-01-06
feat: add ignore_case boolean to ComparisonFilter

kylemumma:
This is for https://github.com/getsentry/eap-planning/issues/127
the next step will be to implemented the functionality for it in the snuba repo

### DIFF
--- a/proto/sentry_protos/snuba/v1/trace_item_filter.proto
+++ b/proto/sentry_protos/snuba/v1/trace_item_filter.proto
@@ -25,6 +25,7 @@ message ComparisonFilter {
     OP_GREATER_THAN_OR_EQUALS = 4;
     OP_EQUALS = 5;
     OP_NOT_EQUALS = 6;
+    OP_EQUALS_IGNORE_CASE = 11; // string only
     OP_LIKE = 7; //string only
     OP_NOT_LIKE = 8; //string only
     OP_IN = 9; // array only
--- a/proto/sentry_protos/snuba/v1/trace_item_filter.proto
+++ b/proto/sentry_protos/snuba/v1/trace_item_filter.proto
@@ -25,7 +25,6 @@ message ComparisonFilter {
     OP_GREATER_THAN_OR_EQUALS = 4;
     OP_EQUALS = 5;
     OP_NOT_EQUALS = 6;
-    OP_EQUALS_IGNORE_CASE = 11; // string only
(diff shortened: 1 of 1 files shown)


proto/sentry_protos/snuba/v1/trace_item_filter.proto
```diff
@@ -25,6 +25,7 @@ message ComparisonFilter {
     OP_GREATER_THAN_OR_EQUALS = 4;
     OP_EQUALS = 5;
     OP_NOT_EQUALS = 6;
+    OP_EQUALS_IGNORE_CASE = 11; // string only
```
phacops, line 28: Could we maybe do a boolean attribute in the `ComparisonFilter` to make the comparison case insensitive? Otherwise we'll need to also implement `IN`, `NOT_IN`, `NOT_EQUALS` with a case insensitive version.

kylemumma: Yeah I can do it that way, I also thought of this issue but couldn't come up with a good way to solve it.

Another thought was something like a `lower()` function on strings but it didnt seem viable with the protobuf definitions

proto/sentry_protos/snuba/v1/trace_item_filter.proto
```diff
@@ -33,6 +33,9 @@ message ComparisonFilter {
   AttributeKey key = 1;
   Op op = 2;
   AttributeValue value = 3;
+
+  // optional, if set will ignore case for string comparisons
+  bool ignore_case = 4;
```
volokluev, line 38: I would prefer you add an `OP_EQUALS_IGNORE_CASE`. This is too confusing

Zylphrex:
Another idea. What if we introduced a new type called `TYPE_STRING_CASE_INSENSITIVE`?

phacops:
> What if we introduced a new type called `TYPE_STRING_CASE_INSENSITIVE`?

I don't think we should tied the insensitiveness to a type, it's should be tied to the comparison op.

xurui-c:
Should we add `ignore_case` boolean to `AggregationComparisonFilter` as well? https://github.com/getsentry/sentry-protos/blob/e62e1b042d00c639eca654c8ac520f8617e34dfc/proto/sentry_protos/snuba/v1/endpoint_trace_item_table.proto#L51

PR #85, opened 2025-01-10
fix(eap-api): Use doubles arrays instead of floats arrays

xurui-c:
https://github.com/getsentry/eap-planning/issues/146

Does https://github.com/getsentry/sentry-protos/pull/84 for arrays as well

### DIFF
--- a/proto/sentry_protos/snuba/v1/trace_item_attribute.proto
+++ b/proto/sentry_protos/snuba/v1/trace_item_attribute.proto
@@ -49,22 +49,30 @@ message IntArray {
   repeated int64 values = 1;
 }
 
+// DEPRECATED, use DoubleArray instead
 message FloatArray {
   repeated float values = 1;
+}
+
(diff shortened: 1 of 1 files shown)


volokluev (approved):

proto/sentry_protos/snuba/v1/trace_item_attribute.proto
```diff
@@ -63,8 +68,9 @@ message AttributeValue {
     bool val_null = 5;
     StrArray val_str_array = 6;
     IntArray val_int_array = 7;
-    FloatArray val_float_array = 8;
+    FloatArray val_float_array = 8 [deprecated = true];
```
kylemumma, line 71: could you add a deprecated comment above this, I dont think it shows that its deprecated in the docs otherwise

kylemumma (approved):

kylemumma (approved):

PR #87, opened 2025-01-13
fix(eap-api): change float type to double type

xurui-c:
https://github.com/getsentry/eap-planning/issues/146

### DIFF
--- a/proto/sentry_protos/snuba/v1alpha/trace_item_attribute.proto
+++ b/proto/sentry_protos/snuba/v1alpha/trace_item_attribute.proto
@@ -7,8 +7,10 @@ message AttributeKey {
     TYPE_UNSPECIFIED = 0; //called "none" sometimes
     TYPE_BOOLEAN = 1;
     TYPE_STRING = 2;
-    TYPE_FLOAT = 3;
+    // deprecated, use TYPE_DOUBLE instead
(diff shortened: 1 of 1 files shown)


proto/sentry_protos/snuba/v1alpha/trace_item_attribute.proto
```diff
@@ -7,8 +7,10 @@ message AttributeKey {
     TYPE_UNSPECIFIED = 0; //called "none" sometimes
     TYPE_BOOLEAN = 1;
     TYPE_STRING = 2;
-    TYPE_FLOAT = 3;
+    // deprecated, use TYPE_DOUBLE instead
+    TYPE_FLOAT = 3 [deprecated = true];
     TYPE_INT = 4; //note: all numbers are stored as float64, so massive integers can be rounded. USE STRING FOR IDS.
+    TYPE_DOUBLE = 6;
```
davidtsuk, line 13: shouldn't this be 5 not 6?

xurui-c: good catch my bad thanks!

davidtsuk (approved):
LGTM

xurui-c:
this PR is wrong because I did it for v1alpha instead of v1

PR #89, opened 2025-01-13
fix(eap): Only allow compraison between aggregation and a double in post aggregation filter

davidtsuk:


proto/sentry_protos/snuba/v1/endpoint_trace_item_table.proto
```diff
@@ -58,9 +58,11 @@ message AggregationComparisonFilter {
     OP_EQUALS = 5;
     OP_NOT_EQUALS = 6;
   }
-  Column column = 1;
+  Column column = 1 [deprecated = true];
```
volokluev, line 61: explain why it's deprecated in comments

volokluev (approved):

PR #94, opened 2025-01-14
chore: Refactor trace_item_name enum into trace_item_type, add log object

colin-sentry:
This will allow querying logs from EAP


proto/sentry_protos/snuba/v1/request_common.proto
```diff
@@ -47,6 +47,7 @@ enum TraceItemName {
   TRACE_ITEM_NAME_UNSPECIFIED = 0;
   TRACE_ITEM_NAME_EAP_SPANS = 1;
   TRACE_ITEM_NAME_EAP_ERRORS = 2;
+  TRACE_ITEM_NAME_OURLOGS = 3;
```
phacops, line 50: I think we should call that `LOG`. I don't know why we went with plural either.

colin-sentry: All of the other types are plural above? We went with a 'brand name' so you could grep for it, 'log' gets infinite false positives due to statements like `console.log`

k-fish: Plural probably comes from the dataset names in sentry. 

For `ourlog` vs `log`, in any user facing code it's fine being `log` but aside from that as Colin mentioned I'd prefer it being more discoverable. It conflates with internal logging and other logs in almost every repo, packages, crates etc. otherwise.

k-fish: `TRACE_ITEM_NAME_LOG` imo is fine though  if we're switching the rest eventually too, it's qualified enough it's not ambiguous.

proto/sentry_protos/snuba/v1/endpoint_find_traces.proto
```diff
@@ -45,8 +46,9 @@ message FindTracesResponse {
 }
 
 message EventFilter {
-  TraceItemName trace_item_name = 1;
+  TraceItemName trace_item_name = 1 [deprecated = true];
```
phacops, line 49: Let's remove those changes, it's not even implemented.

proto/sentry_protos/snuba/v1/endpoint_get_trace.proto
```diff
@@ -12,9 +12,11 @@ import "sentry_protos/snuba/v1/trace_item_attribute.proto";
 // attributes to return for each item type.
 message GetTraceRequest {
   message TraceItem {
+    //DEPRECATED: use item_type instead
     TraceItemName type = 1;
```
phacops, line 16: Why don't you add the `deprecated` annotation here too?

colin-sentry: done

phacops (approved):

PR #97, opened 2025-01-20
ref(taskbroker): Simplify task activation and retry state

enochtangg:
Create a new v2 `taskbroker.proto` from v1 `taskworker.proto` and refactor its schema. 

Changes:
* Created `OnAttemptsExceeded` enum to capture what happens to task after attempts are exceeded. In proto3, explicit enum default values are not supported. Therefore, we cannot default to `ON_ATTEMPTS_EXCEEDED_DISCARD`. Instead, the default is automatically set to `ON_ATTEMPTS_EXCEEDED_UNSPECIFIED`. In our application, we need to handle this as discard.
* Simplify `RetryState` to use `max_attempts` and `on_attempts_exceeded` enum. I prefer `max_attempts` over `attempts_remaining` since the latter implies that it's a decrementing counter.
* Change RetryState attributes `int32` -> `unint32`
* Removed `discard_after_attempt` and `deadletter_after_attempt` attributes.
* Removed un-used `InflightActivation` message type.
* Removed deprecated attributes.

### DIFF
--- a/proto/sentry_protos/sentry/v2/taskbroker.proto
+++ b/proto/sentry_protos/sentry/v2/taskbroker.proto
@@ -13,7 +13,7 @@ enum OnAttemptsExceeded {
 
 message RetryState {
   // Current attempt number
-  int32 attempts = 1;
+  uint32 attempts = 1;
 
   // The classname or adapter type for the retry policy
   string kind = 2;
--- a/proto/sentry_protos/sentry/v2/taskbroker.proto
+++ b/proto/sentry_protos/sentry/v2/taskbroker.proto
@@ -1,9 +1,15 @@
 syntax = "proto3";
 
-package sentry_protos.sentry.v1;
+package sentry_protos.sentry.v2;
 
 import "google/protobuf/any.proto";
 import "google/protobuf/timestamp.proto";
+
+enum OnAttemptsExceeded {
+  ON_ATTEMPTS_EXCEEDED_UNSPECIFIED = 0;
+  ON_ATTEMPTS_EXCEEDED_DISCARD = 1;
+  ON_ATTEMPTS_EXCEEDED_DEADLETTER = 2;
+}
 
 message RetryState {
   // Current attempt number
@@ -12,11 +18,11 @@ message RetryState {
   // The classname or adapter type for the retry policy
   string kind = 2;
 
-  // After this attempt the task should be discarded
-  optional int32 discard_after_attempt = 3;
+  // After this number of attempts, the task is either discarded or deadlettered.
+  uint32 max_attempts = 3;
 
-  // After this attempt the task should be put in the dead-letter-queue.
-  optional int32 deadletter_after_attempt = 4;
+  // The action to take after the attempts_limit is exceeded.
+  OnAttemptsExceeded on_attempts_exceeded = 4;
 
   // Whether a task should be executed at most once.
   optional bool at_most_once = 5;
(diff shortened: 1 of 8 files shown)


proto/sentry_protos/sentry/v2/taskbroker.proto
```diff
@@ -0,0 +1,124 @@
...
+  ON_ATTEMPTS_EXCEEDED_DISCARD = 1;
+  ON_ATTEMPTS_EXCEEDED_DEADLETTER = 2;
+}
+
+message RetryState {
+  // Current attempt number
+  uint32 attempts = 1;
+
+  // The classname or adapter type for the retry policy
+  string kind = 2;
```
markstory, line 19: We haven't found a scenario where we need this yet. Until we have that scenario we could omit this attribute. Adding new attributes in the future is simple enough.

proto/sentry_protos/sentry/v2/taskbroker.proto
```diff
@@ -0,0 +1,124 @@
...
+
+  // After this number of attempts, the task is either discarded or deadlettered.
+  uint32 max_attempts = 3;
+
+  // The action to take after the attempts_limit is exceeded.
```
markstory, line 24:
```suggestion
  // The action to take after the max_attempts is exceeded.
```

proto/sentry_protos/sentry/v2/taskbroker.proto
```diff
@@ -0,0 +1,124 @@
...
+enum TaskActivationStatus {
+  TASK_ACTIVATION_STATUS_UNSPECIFIED = 0;
+  TASK_ACTIVATION_STATUS_PENDING = 1;
+  TASK_ACTIVATION_STATUS_PROCESSING = 2;
+  TASK_ACTIVATION_STATUS_FAILURE = 3;
+  TASK_ACTIVATION_STATUS_RETRY = 4;
+  TASK_ACTIVATION_STATUS_COMPLETE = 5;
+}
+
+// Task message that is stored in Kafka.
```
markstory, line 40:
```suggestion
// Task message that is stored in Kafka and shared over RPC
```

proto/sentry_protos/sentry/v2/taskbroker.proto
```diff
@@ -0,0 +1,124 @@
...
+// Once consumed, TaskActivations are wrapped with InflightActivation to track
+// additional state
```
markstory, line 42: I don't know if we need to mention `InflightActivation` here as that structure is only internal to taskbroker right now.

proto/sentry_protos/sentry/v2/taskbroker.proto
```diff
@@ -0,0 +1,124 @@
...
+// RPC messages and services
+////////////////////////////
+message Error {
+  // Taken directly from the grpc docs.
+  int32 code = 1;
+  string message = 2;
+
+  // A list of messages that carry any error details.
+  repeated google.protobuf.Any details = 3;
+}
```
markstory, line 85: I don't think we ended up using this Error struct. Instead we use the protobuf response/status code structs. eg `Status::not_found()`

proto/sentry_protos/sentry/v2/taskbroker.proto
```diff
@@ -0,0 +1,124 @@
...
+
+message GetTaskRequest {
+  optional string namespace = 1;
+}
+
+message GetTaskResponse {
+  // If there are no tasks available, this will be empty
+  optional TaskActivation task = 1;
+
+  optional Error error = 2;
```
markstory, line 103: We could omit this and the other `error` attributes.

rust/src/lib.rs
markstory, line 1: This file generally only changes when releases are made and by accident.

proto/sentry_protos/sentry/v2/taskbroker.proto
markstory, line 1: Should we put this in `sentry_protos/taskbroker/v2/taskbroker.proto` 🤷

enochtangg: yeah let's do that. I started from v1 again

rust/src/lib.rs
enochtangg: It seems my vscode is doing a `cargo check` and re-ordering those modules on save.

markstory (approved):
Looks good to me.

PR #103, opened 2025-01-24
fix(eap-api): add sample count to DataPoint for time series

xurui-c:
https://github.com/getsentry/eap-planning/issues/156

### DIFF
--- a/proto/sentry_protos/snuba/v1/endpoint_time_series.proto
+++ b/proto/sentry_protos/snuba/v1/endpoint_time_series.proto
@@ -57,6 +57,9 @@ message DataPoint {
 
   // the reliability of the data value based on math based on confidence intervals and sample size
   Reliability reliability = 5;
+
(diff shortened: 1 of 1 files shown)


davidtsuk (commented):
I think we should add a comment to describe the new field

kylemumma:
I agree with david

proto/sentry_protos/snuba/v1/endpoint_time_series.proto
```diff
@@ -57,6 +57,9 @@ message DataPoint {
 
   // the reliability of the data value based on math based on confidence intervals and sample size
   Reliability reliability = 5;
+
+  // the extrapolated sample count for this data point
```
davidtsuk, line 61: the sample count shouldn't be extrapolated, it's the actual count

davidtsuk (approved):

PR #104, opened 2025-01-24
feat(eap): Add an is_null field for to AttributeValue

davidtsuk:
This is used to distinguish between fields that are default values (e.g. 0 and empty string) and fields that are null (e.g. when aggregating over an empty set).

The two approaches I thought of were:
1. add an array to represent if the result is null for each column (first commit)
2. add a bool value to the AttributeValue to represent nullity (most recent commit)

Since the first approach would increase the message size, I prefer the second approach which doesn't affect the message size.

### DIFF
--- a/proto/sentry_protos/snuba/v1/endpoint_trace_item_table.proto
+++ b/proto/sentry_protos/snuba/v1/endpoint_trace_item_table.proto
@@ -88,6 +88,10 @@ message TraceItemColumnValues {
 
   // reliability of the values based on confidence interval and sample size math
   repeated Reliability reliabilities = 3;
+
+  // true if the value is present, false if it is not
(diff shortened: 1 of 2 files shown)


xurui-c (approved):

kylemumma:
Is there a reason you chose this over using `Optional`

kylemumma (commented):
Is there a reason you chose this over using `Optional`?

davidtsuk:
> Is there a reason you chose this over using Optional?
@kylemumma It doesn't look like it's possible to use the optional keyword on a repeated field like this

kylemumma (approved):

kylemumma (approved):
approved with the condition that u get pierre or volo also to approve

phacops (approved):

PR #108, opened 2025-02-03
chore: add label to timeseries expression

kylemumma:
I forgot to add this in https://github.com/getsentry/sentry-protos/pull/105 but i need it for timeseries formulas


volokluev (approved):
update the tests please

PR #111, opened 2025-02-20
feat: Add a "trace item details" endpoint to get everything for a single item

colin-sentry:
This lets you do the equivalent of `SELECT * FROM trace_items WHERE item_type=log AND item_id=12345`

### DIFF
--- a/proto/sentry_protos/snuba/v1/endpoint_trace_item_details.proto
+++ b/proto/sentry_protos/snuba/v1/endpoint_trace_item_details.proto
@@ -26,7 +26,7 @@ message TraceItemDetailsAttribute {
 // this is a response from the TraceItemDetails endpoint
 message TraceItemDetailsResponse {
   string item_id = 1;
-  google.protobuf.Timestamp start_timestamp = 2;
+  google.protobuf.Timestamp timestamp = 2;
   repeated TraceItemDetailsAttribute attributes = 3;
   ResponseMeta meta = 4;
 }
--- a/proto/sentry_protos/snuba/v1/endpoint_trace_item_details.proto
+++ b/proto/sentry_protos/snuba/v1/endpoint_trace_item_details.proto
@@ -1,0 +1,29 @@
+syntax = "proto3";
+
+package sentry_protos.snuba.v1;
+
+import "sentry_protos/snuba/v1/request_common.proto";
+import "sentry_protos/snuba/v1/trace_item_attribute.proto";
+
+// This is a request to the TraceItemDetails endpoint,
+// it is used to query for TraceItems (ex. spans or logs)
+// it returns all of the information for a specific trace item.
+message TraceItemDetailsRequest {
+  RequestMeta meta = 1;
+
+  //required: the UUID of the item you are looking for
+  string item_uuid = 2;
+  //OPTIONAL: improves performance if included
+  string trace_id = 3;
+}
+
+message TraceItemDetailsAttribute {
+  string name = 1;
+  AttributeValue value = 2;
+}
+
+// this is a response from the TraceItemTable endpoint
+// it is the counterpart to TraceItemTableRequest
+message TraceItemDetailsResponse {
+  repeated TraceItemDetailsAttribute attributes = 1;
+}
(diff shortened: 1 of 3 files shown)


py/tests/test_snuba_v1.py
```diff
@@ -381,6 +387,23 @@ def test_example_table_with_aggregation_filter() -> None:
         ), 
     )
 
+
+def text_trace_item_details() -> None:
```
mjq, line 391: Did you mean
```suggestion
def test_trace_item_details() -> None:
```
?

proto/sentry_protos/snuba/v1/endpoint_trace_item_details.proto
```diff
@@ -0,0 +1,29 @@
...
+import "sentry_protos/snuba/v1/trace_item_attribute.proto";
+
+// This is a request to the TraceItemDetails endpoint,
+// it is used to query for TraceItems (ex. spans or logs)
+// it returns all of the information for a specific trace item.
+message TraceItemDetailsRequest {
+  RequestMeta meta = 1;
+
+  //required: the UUID of the item you are looking for
+  string item_uuid = 2;
```
k-fish, line 15: How does this work with span_id being 64?

colin-sentry: hex string now, fix't

proto/sentry_protos/snuba/v1/endpoint_trace_item_details.proto
```diff
@@ -0,0 +1,29 @@
...
+// This is a request to the TraceItemDetails endpoint,
+// it is used to query for TraceItems (ex. spans or logs)
+// it returns all of the information for a specific trace item.
+message TraceItemDetailsRequest {
+  RequestMeta meta = 1;
+
+  //required: the ID (hex string) of the item you are looking for
+  string item_id = 2;
+  //OPTIONAL: improves performance if included
+  string trace_id = 3;
```
k-fish, line 17: Should this be the filter as per the notion https://www.notion.so/sentry/EAP-API-Design-1198b10e4b5d80faa5e9dcec47a9c12b?pvs=4#11b8b10e4b5d80e78ad8f13f9e58c9ff 
```suggestion
  repeated TraceItemFilter filter = 3;
```

proto/sentry_protos/snuba/v1/endpoint_trace_item_details.proto
```diff
@@ -0,0 +1,29 @@
...
+}
+
+message TraceItemDetailsAttribute {
+  string name = 1;
+  AttributeValue value = 2;
+}
+
+// this is a response from the TraceItemTable endpoint
+// it is the counterpart to TraceItemTableRequest
```
phacops, line 26: Not sure what this means?

proto/sentry_protos/snuba/v1/endpoint_trace_item_details.proto
```diff
@@ -0,0 +1,29 @@
...
+message TraceItemDetailsResponse {
+  repeated TraceItemDetailsAttribute attributes = 1;
```
phacops, line 28: I think we should repeat `item_id`/`timestamp`/`trace_id` as fields on this since they are mandatory and the rest would go to attributes.

proto/sentry_protos/snuba/v1/endpoint_trace_item_details.proto
phacops: Let's indeed leave it as a filter. If we make it required later only, we'll promote it.

proto/sentry_protos/snuba/v1/endpoint_trace_item_details.proto
colin-sentry: disagree with this, it just makes consuming this clunkier - those attributes will almost always go to the same place, regardless of if they are 'guaranteed to exist' or not

phacops:
Add the responsemeta in the response as well please.

proto/sentry_protos/snuba/v1/endpoint_trace_item_details.proto
phacops: What would make it clunkier is to search through a list of attributes to get the item ID when we know it exists and always returned.

phacops (approved):

PR #112, opened 2025-02-25
feat(smart-autocomplete): make api changes for smart autocomplete

volokluev:
Context here:

https://www.notion.so/sentry/Smart-autocomplete-18a8b10e4b5d80a9967aca8f8fd7a598

closes https://github.com/getsentry/eap-planning/issues/198


rust/src/lib.rs
```diff
@@ -1,30 +1,30 @@
-pub mod seer {
+pub mod events {
```
onkar, line 1: Are the changes in this file intended? Seems like something re-ordered things here.

onkar (approved):
LGTM except the changes in `rust/src/lib.rs`. If they are not intended, I'd recommend reverting that file.

PR #113, opened 2025-03-12
feat(eap): Add schema for TraceItem

phacops:


proto/sentry_protos/snuba/v1/request_common.proto
```diff
@@ -53,14 +54,6 @@ enum TraceItemName {
...
 }
 
-enum TraceItemType {
-  TRACE_ITEM_TYPE_UNSPECIFIED = 0;
-  TRACE_ITEM_TYPE_SPAN = 1;
-  TRACE_ITEM_TYPE_ERROR = 2;
-  TRACE_ITEM_TYPE_LOG = 3;
-  TRACE_ITEM_TYPE_UPTIME_CHECK = 4;
-}
-
```
volokluev, line 63: I think this breaks backwards compatibility

proto/sentry_protos/snuba/v1/trace_item.proto
```diff
@@ -0,0 +1,28 @@
...
+  TRACE_ITEM_TYPE_ERROR = 2;
+  TRACE_ITEM_TYPE_LOG = 3;
+  TRACE_ITEM_TYPE_UPTIME_CHECK = 4;
+}
+
+message TraceItem {
+  uint64 organization_id = 1;
+  uint64 project_id = 2;
+  string trace_id = 3;
+  uint64 item_id = 4;
```
volokluev, line 20: don't we store item_id as a Uint128?

rust/build.rs
```diff
@@ -65,66 +52,10 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     println!("Generating proto bindings");
     tonic_build::configure()
         .emit_rerun_if_changed(false)
+        .include_file("./include.rs")
```
volokluev, line 55: beautiful

proto/sentry_protos/snuba/v1/trace_item.proto
phacops: We do, we do, I need to change that, I wanted to make something that was generating the protos. I think I'll replace that with a `bytes` type since we don't have `u128`.

proto/sentry_protos/snuba/v1/request_common.proto
phacops: It does, I'll revert.

markstory (commented):
I tried these changes out locally, and couldn't get the local dev flow that uses
```
sentry_protos = "0.1.0"

[patch.crates-io]
sentry_protos = { path = "../sentry-protos/rust/" }
```

in a rust application to compile. I got compiler errors that proto modules weren't defined.

proto/sentry_protos/snuba/v1/trace_item.proto
```diff
@@ -0,0 +1,22 @@
...
+
+message TraceItem {
+  uint64 organization_id = 1;
+  uint64 project_id = 2;
+  string trace_id = 3;
+  // item_id as bytes, little endian
+  bytes item_id = 4;
+  TraceItemType item_type = 5;
+  google.protobuf.Timestamp timestamp = 6;
+  repeated opentelemetry.proto.common.v1.KeyValue attributes = 7;
```
k-fish, line 17: I know this is only a proto in snuba for now, but are we thinking of making the EAP kafka also use attributes for consistency?

proto/opentelemetry/proto/common/v1/common.proto
```diff
@@ -0,0 +1,81 @@
...
+option java_multiple_files = true;
+option java_outer_classname = "CommonProto";
+option java_package = "io.opentelemetry.proto.common.v1";
+
+// AnyValue is used to represent any type of attribute value. AnyValue may contain a
+// primitive value such as a string or integer or it may contain an arbitrary nested
+// object containing arrays, key-value lists and primitives.
+message AnyValue {
+  // The value is one of the listed fields. It is valid for all values to be unspecified
+  // in which case this AnyValue is considered to be "empty".
```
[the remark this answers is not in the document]
phacops: No, we can't do that right now. We don't store the key if we need a `null` value.

phacops:
> What do you want the module paths for these opentelemetry protos to be?

I'm looking for `opentelemetry::proto::common::v1::KeyValue` to mirror the `opentelemetry-rust` crate.

But the way it prefixes by `sentry_protos` twice, while working, is not great. I'm sure we can tweak the behavior.

markstory (approved):

volokluev (approved):

PR #117, opened 2025-04-09
feat(taskworker): add optional delay value to ktasks

john-z-yang:


proto/sentry_protos/sentry/v1/taskworker.proto
```diff
@@ -67,6 +67,10 @@ message TaskActivation {
   // The duration in seconds that a task has to start execution.
   // After received_at + expires has passed an activation is expired and will not be executed.
   optional uint64 expires = 10;
+
+  // The duration in seconds that a task can start execution after it is emitted.
+  // After received_at + delay has passed an activation is will be executed.
```
markstory, line 72:
```suggestion
  // The duration in seconds that a task must wait to begin execution after it is emitted.
  // After received_at + delay has passed, the activation will become pending.
```

markstory (approved):

PR #119, opened 2025-04-14
feat: default value for formulas

kylemumma:
this allows user to specify default values for formulas in the traceitemtable and timeseries endpoints.

still needs support for it to be implemented in snuba.

it is for this ticket
https://github.com/orgs/getsentry/projects/284/views/1?pane=issue&itemId=104071496&issue=getsentry%7Ceap-planning%7C232

### DIFF
--- a/proto/sentry_protos/snuba/v1/endpoint_time_series.proto
+++ b/proto/sentry_protos/snuba/v1/endpoint_time_series.proto
@@ -65,7 +65,8 @@ message Expression {
     Op op = 1;
     Expression left = 2;
     Expression right = 3;
-    // optional, if the formula is invalid, use this value
+    // optional, if the formula evalutates to null this value is returned
+    // formula can evaluate to null if an attribute is missing for example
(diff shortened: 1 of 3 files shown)


proto/sentry_protos/snuba/v1/endpoint_time_series.proto
```diff
@@ -65,6 +65,11 @@ message Expression {
     Op op = 1;
     Expression left = 2;
     Expression right = 3;
+    // optional, if the formula is invalid, use this value
```
volokluev, line 68: why would the formula be invalid?

kylemumma: it happens if an attribute is missing for example `div(a,b)` but some item doesnt have both `a` and `b` then itll evaluate to null and the default value is used.

I updated the comment to be more clear, and linked the original ticket with context in the pr.

volokluev (approved):

PR #120, opened 2025-04-22
feat(storage-routing): NORMAL mode

xurui-c:


proto/sentry_protos/snuba/v1/downsampled_storage.proto
```diff
@@ -7,6 +7,7 @@ message DownsampledStorageConfig {
     MODE_UNSPECIFIED = 0;
     MODE_PREFLIGHT = 1;
     MODE_BEST_EFFORT = 2;
```
volokluev, line 9: add a comment saying these two are deprecated and that they map to normal mode

proto/sentry_protos/snuba/v1/downsampled_storage.proto
```diff
@@ -7,6 +7,7 @@ message DownsampledStorageConfig {
...
+    MODE_NORMAL = 3;
```
volokluev, line 10: explain what this is

volokluev (approved):

PR #126, opened 2025-05-07
feat(taskbroker): Add option to delay on retry

john-z-yang:


proto/sentry_protos/taskbroker/v1/taskbroker.proto
```diff
@@ -68,6 +68,9 @@ message TaskActivation {
   // The duration in seconds that a task must wait to begin execution after it is emitted.
   // After received_at + delay has passed, the activation will become pending.
   optional uint64 delay = 11;
+
+  // Duration in seconds that a task must wait to begin execution after it is retried.
+  optional uint64 delay_on_retry = 12;
```
markstory, line 73: Could we put this into `RetryState`?  That will let us do `Retry(times=3, delay=30)` in applications that declare tasks.

john-z-yang: Yeah sure

proto/sentry_protos/taskbroker/v1/taskbroker.proto
```diff
@@ -22,6 +22,9 @@ message RetryState {
 
   // Whether a task should be executed at most once.
   optional bool at_most_once = 4;
+
+  // Duration in seconds that a task must wait to begin execution after it is retried.
+  optional uint64 delay_on_retry = 5;
```
evanh, line 27: If it's in the retry state, I would just call it delay?

john-z-yang: I think I prefer to be more explicit here since we already have another `delay` does something completely different

markstory (approved):
Looks good to me.

PR #127, opened 2025-05-07
feat: Generate Protobuf files for release

phacops:
As we integrated this in Relay, we realize imposing the dependency of `protoc` when building the `processing` feature was too much.

Instead, this will allow to generate the Rust representation of the Protobuf files when we release the crate.


.github/workflows/release.yml
```diff
@@ -27,6 +27,20 @@ jobs:
...
+      - name: Generate rust code
+        run: make build-rust
+
+      - name: Commit generated rust code
+        run: |
+          git config user.email bot@getsentry.com
+          git config user.name getsentry-bot
+          git add rust/src
+          git add Cargo.lock
+          git commit -m 'chore: Regenerate Rust bindings'
```
Dav1dde, line 42: As mentioned on Slack, this might fail due to a empty commit.

Dav1dde (commented):
Also mentioned on slack (as an idea):

Instead of generating the files on release, that could run in the scope of the PR to get instant diffs and it would enable depending on a git dependency of the crate for testing. Also maybe just running it as a check is fine, then there won't be an annoying CI job that commits to your branches.

build_rust/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Installation
+
+```
+cargo add sentry_protos
+```
+
+## Generating bindings
+
+Rust bindings are generated with the `rustgenerator` package. Bindings can be generated using
```
markstory, line 14:
```suggestion
Rust bindings are generated with the `build_sentry_protos` package. Bindings can be generated using
```

build_rust/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+```shell
+# In root of the repository
+make build-rust
+
+# Or with
+cargo run -p rustgenerator
```
markstory, line 21:
```suggestion
cargo run -p build_sentry_protos
```

markstory:
> Instead of generating the files on release, that could run in the scope of the PR to get instant diffs and it would enable depending on a git dependency of the crate for testing. Also maybe just running it as a check is fine, then there won't be an annoying CI job that commits to your branches.

I agree that having generated code added to pull requests would be a better workflow as you can see what you're going to get in terms of generated rust code. We could do that as a follow up change one Pierre is unblocked though.

Dav1dde (approved):
Agree with Mark, we can always improve the workflow in a follow-up 👍 

Thanks for taking care of this.

markstory (approved):

PR #128, opened 2025-06-12
ref(eap): allow label for item table orderby

MeredithAnya:
Addresses [EAP-90](https://linear.app/getsentry/issue/EAP-90/improve-order-by-in-protobuf-requests)

`OrderBy` can now additionally allow the order by to be the label directly instead of the full `Column` definition (which ultimately uses `column.label` for orderby)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -764,7 +764,7 @@ checksum = "955d28af4278de8121b7ebeb796b6a45735dc01436d898801014aced2773a3d6"
 
 [[package]]
 name = "sentry_protos"
-version = "0.1.75"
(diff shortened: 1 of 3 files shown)


MeredithAnya:
> I don't think you need to do this. All fields in protobuf are optional. So a user should just be able to specify `Column(label="my_label")` and that should work fine

Okays so then this improvement is only requires a snuba change then? is that what you are saying?

MeredithAnya:
closing since we don't actually need this - https://github.com/getsentry/snuba/pull/7236 shows that we only need to make changes in snuba not here

PR #133, opened 2025-07-28
feat(eap): add BinaryFormula to AggregationComparisonFilter

onewland:
Reusing Column.BinaryFormula here because creating a common BinaryFormula was putting me into dependency hell. But open to other approaches 

https://linear.app/getsentry/issue/EAP-74/allow-aggregationcomparisonfilter-to-support-binaryformula

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,2 +1,2 @@
-/proto/sentry_protos/snuba                       @getsentry/performance @getsentry/owners-snuba
+/proto/sentry_protos/snuba                       @getsentry/owners-snuba
 /proto/sentry_protos/taskbroker                  @getsentry/taskbroker
--- a/proto/sentry_protos/snuba/v1/trace_item.proto
+++ b/proto/sentry_protos/snuba/v1/trace_item.proto
@@ -46,4 +46,5 @@ message TraceItem {
   // Internal fields
   uint32 retention_days = 100;
   google.protobuf.Timestamp received = 101;
+  uint32 downsampled_retention_days = 102;
 }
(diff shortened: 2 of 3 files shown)


VERSION
```diff
@@ -1 +1 @@
-0.2.0
```
kylemumma, line 1: whats this for? i always bump the version using the github action

rust/src/sentry_protos.snuba.v1.rs
```diff
@@ -1626,6 +1626,8 @@ pub struct AggregationComparisonFilter {
     pub aggregation: ::core::option::Option<AttributeAggregation>,
     #[prost(message, optional, tag = "6")]
     pub conditional_aggregation: ::core::option::Option<AttributeConditionalAggregation>,
+    #[prost(message, optional, tag = "7")]
+    pub formula: ::core::option::Option<column::BinaryFormula>,
```
kylemumma, line 1630: whats this file? ive never changed this file when making a protobuf change

VERSION
onewland: I couldn't get it to load locally with `pip install -e ../<path>`. But if this will automatically update on merge, I'll remove this change

rust/src/sentry_protos.snuba.v1.rs
onewland: I think I ran these commands and committed the changes, but I can also back that out if we want https://github.com/getsentry/sentry-protos?tab=readme-ov-file#rust

rust/src/sentry_protos.snuba.v1.rs
```diff
@@ -1913,4 +1916,6 @@ pub struct TraceItem {
     pub retention_days: u32,
     #[prost(message, optional, tag = "101")]
     pub received: ::core::option::Option<::prost_types::Timestamp>,
+    #[prost(uint32, tag = "102")]
+    pub downsampled_retention_days: u32,
```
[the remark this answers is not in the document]
onewland: merging from main removed this part of the diff

PR #135, opened 2025-07-29
feat(eap): Allow to pass a different retention for downsampled data

phacops:
This PR allows us to pass a different value for retention of the downsampled datasets.


proto/sentry_protos/snuba/v1/trace_item.proto
```diff
@@ -46,4 +46,5 @@ message TraceItem {
   // Internal fields
   uint32 retention_days = 100;
   google.protobuf.Timestamp received = 101;
+  uint32 downsampled_retention_days = 102;
```
onkar, line 49: nit: can we add a comment on the difference between `retention_days` and `downsampled_retention_days`?

onkar (approved):
Looks pretty straightforward.

PR #138, opened 2025-09-03
feat: heatmap support for traceitemstats endpoint

kylemumma:
this is the interface definition for this https://linear.app/getsentry/issue/EAP-190/support-heat-map-representation-in-traceitemstats-endpoint

see here for example usage https://www.notion.so/sentry/example-request-and-response-2638b10e4b5d801f9ff8ebaaab9ff970

## design decisions for the endpoint
- bucket ranges for the numerical attribute will be the same across all attributes and values based on the global range
- the highest occurring attributes are the ones shown
- the highest occurring values within attributes are the ones shown

### DIFF
--- a/proto/sentry_protos/snuba/v1/endpoint_trace_item_stats.proto
+++ b/proto/sentry_protos/snuba/v1/endpoint_trace_item_stats.proto
@@ -11,9 +11,17 @@ message AttributeDistributionsRequest {
   uint32 max_attributes = 2;
 }
 
+message HeatmapRequest {
+  uint32 max_attribute_value_buckets = 1;
+  uint32 max_attributes = 2;
+  uint32 num_heatmap_buckets = 3;
+  string numerical_grouping_attribute_name = 4;
+}
+
 message StatsType {
   oneof type {
     AttributeDistributionsRequest attribute_distributions = 1;
+    HeatmapRequest heatmap = 2;
   }
 }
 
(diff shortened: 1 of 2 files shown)


proto/sentry_protos/snuba/v1/endpoint_trace_item_stats.proto
```diff
@@ -11,9 +11,17 @@ message AttributeDistributionsRequest {
   uint32 max_attributes = 2;
 }
 
+message HeatmapRequest {
+  uint32 max_attribute_value_buckets = 1;
+  uint32 max_attributes = 2;
+  uint32 num_heatmap_buckets = 3;
+  string numerical_grouping_attribute_name = 4;
+}
+
```
volokluev, line 20: Would this make more sense to make it its own endpoint? it seems like the request and results are both`oneOf`s

proto/sentry_protos/snuba/v1/endpoint_trace_item_stats.proto
```diff
@@ -11,9 +11,17 @@ message AttributeDistributionsRequest {
   uint32 max_attributes = 2;
 }
 
+message HeatmapRequest {
+  uint32 max_attribute_value_buckets = 1;
+  uint32 max_attributes = 2;
+  uint32 num_heatmap_buckets = 3;
+  string numerical_grouping_attribute_name = 4;
```
volokluev, line 18: what do these parameters represent? It's not obvious to me as a user of the API

proto/sentry_protos/snuba/v1/endpoint_trace_item_stats.proto
kylemumma: I was debating about this as well but I decided to keep it in the same endpoint since:
* the new protobuf definition is an easy change
* a new endpoint would likely have a lot of duplicated code

with the downside being
* users will have to validate the oneof in the response even though you know which one is will be 

lmk if you feel strongly one way

proto/sentry_protos/snuba/v1/endpoint_trace_item_stats.proto
kylemumma:
```
span.op = "db.query":
  - duration [0-300ms): 400 occurrences
  - duration [300-600ms): 600 occurrences  
  - duration [600-900ms): 300 occurrences
  - duration [900-1200ms): 150 occurrences
  - duration [1200-1500ms): 50 occurrences

span.op = "http.request":
  - duration [0-300ms): 200 occurrences
  - duration [300-600ms): 300 occurrences
  - duration [600-900ms): 200 occurrences
  - duration [900-1200ms): 80 occurrences
  - duration [1200-1500ms): 20 occurrences
```
in this example response:
* max_attributes = 1
* max_attribute_value_buckets = 2
* num_heatmap_buckets = 5
* numerical_grouping_attribute_name = duration

PR #141, opened 2025-09-15
feat(cbrs): add time boundaries in response

xurui-c:
https://www.notion.so/sentry/Logs-Query-Routing-2628b10e4b5d803b9bb4ef96af55725c?source=copy_link#2698b10e4b5d80aebb58d07ba2a40b77

With the HIGHEST_ACCURACY_FLEXTIME mode, we send these log queries to tier 1 but in order to meet the time deadline (30s), we truncate start/end time of log queries. So we have to inform the user how we truncated it


volokluev:
not needed rn

PR #144, opened 2025-09-17
add DeleteTrace endpoint request/response

onewland:
Add an interface for deleting all items in a trace (requested by trace ID)

https://linear.app/getsentry/issue/EAP-282/create-new-rpc-endpoint-for-deletion-from-eap-items


proto/sentry_protos/snuba/v1/endpoint_delete_trace.proto
```diff
@@ -0,0 +1,22 @@
...
+
+// DeleteTraceRequest initiates a delete of all items in a trace. It
+// should also remove downsampled copies.
+//
+// The response will be an acknowledgement of a request, but not
+// a guarantee that the delete has been fulfilled.
+message DeleteTraceRequest {
+  RequestMeta meta = 1;
+  string trace_id = 2;
+}
```
volokluev, line 15: I think that it would be better to make a DeleteItemsRequest that takes a TraceItemFilter and a trace_id, so you don't need to write two different endpoints

phacops: And in the spirit of a batch delete, make a `repeated trace_ids`.

onewland:
@phacops @volokluev I believe I've addressed all comments

proto/sentry_protos/snuba/v1/endpoint_delete_trace_items.proto
```diff
@@ -0,0 +1,25 @@
...
+
+// DeleteTraceItemsRequest initiates a delete of all matching items in a trace.
+// It should also remove downsampled copies.
+//
+// The response will be an acknowledgement of a request, but not
+// a guarantee that the delete has been fulfilled.
+message DeleteTraceItemsRequest {
+  RequestMeta meta = 1;
+  repeated string trace_ids = 2;
+  TraceItemFilterWithType filter = 3;
```
phacops, line 16: I think this should be a repeated field as we may need to specify several item types with conditions.

proto/sentry_protos/snuba/v1/endpoint_delete_trace_items.proto
```diff
@@ -0,0 +1,25 @@
+syntax = "proto3";
+
+package sentry_protos.snuba.v1;
+
+import "sentry_protos/snuba/v1/request_common.proto";
+// import "sentry_protos/snuba/v1/trace_item_filter.proto";
+
+// DeleteTraceItemsRequest initiates a delete of all matching items in a trace.
+// It should also remove downsampled copies.
```
phacops, line 9: Can you specify we will either delete the `trace_ids` or the `filter` conditions, but not both at the same time?

onewland:
I think this now takes all feedback into account

phacops (approved):

PR #197, opened 2026-04-14
Add new billing data categories

noahsmartin:
Add data categories for transaction processed/indexed, profile indexed, metric buckets, attachment items, log items, prevent user, seer user, backend/UI profiles, and trace metric bytes.

### DIFF
--- a/proto/sentry_protos/billing/v1/data_category.proto
+++ b/proto/sentry_protos/billing/v1/data_category.proto
@@ -51,6 +51,24 @@ enum DataCategory {
   DATA_CATEGORY_PROFILE_CHUNK_UI = 20;
   // Spans stored in full (indexed).
   DATA_CATEGORY_SPAN_INDEXED = 21;
+  // Processed transaction events.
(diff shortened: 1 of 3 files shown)


proto/sentry_protos/billing/v1/data_category.proto
```diff
@@ -51,6 +51,28 @@ enum DataCategory {
...
+  // Metric buckets.
+  DATA_CATEGORY_METRIC_BUCKET = 25;
+  // Individual attachment items.
+  DATA_CATEGORY_ATTACHMENT_ITEM = 26;
+  // Individual log items.
+  DATA_CATEGORY_LOG_ITEM = 27;
+  // Prevent user events.
+  DATA_CATEGORY_PREVENT_USER = 28;
+  // Seer user events.
+  DATA_CATEGORY_SEER_USER = 29;
```
brendanhsentry: We should be able to remove these prevent user should be deprecated / never used in the first place and seer users are seat based

brendanhsentry (approved):

PR #198, opened 2026-04-14
feat: Add ifNull function

wmak:
- This adds an iffNull function so we can do arithmetic on values that might be sparse


kylemumma (approved):

wmak:
oh hm there's no way to chain functions. gonna look into this further

PR #200, opened 2026-04-17
feat(billing): Add invoicer service and uninvoiced contracts endpoint

noahsmartin:
Introduce a new Invoicer service with a CreateInvoices endpoint that accepts a current timestamp and is intended to be invoked on a schedule to generate invoices for contracts whose billing period has ended.

Add a matching GetUninvoicedContracts endpoint on the Contract service so the Invoicer can enumerate contracts that need invoicing for the current period.


proto/sentry_protos/billing/v1/services/contract/v1/endpoint_get_uninvoiced_contracts.proto
```diff
@@ -0,0 +1,16 @@
...
+
+message GetUninvoicedContractsRequest {
+  // Returns Contracts whose current billing period ends before this time and
+  // have not yet been invoiced for the current period.
+  google.protobuf.Timestamp current_ts = 1;
+}
+
+message GetUninvoicedContractsResponse {
+  repeated Contract contracts = 1;
+}
```
volokluev, line 25: I am pretty sure this will need to be paginated

noahsmartin: Good call. I wonder if what we should do is just have a max_items in the request and since the handling of the response will invoice the contracts we can then call the same endpoint again and we'll get new (uninvoiced) contracts. So not true pagination but easier and same result

volokluev (approved):

PR #201, opened 2026-04-17
feat(billing-platform): package service protos

volokluev:
This PR creates a package service directory and extracts the PricingConfig out of the V1 contract service and puts it onto the package service. 

In doing so, I'm pulling out things which are going to be used across services such as:

* Tier definitions
* billable metrics
* LineItems

I am also moving AWAY from defining SKUs in protos and away from the name SKU in general. It has too many connotations, nobody knows what it means. The word we are using to express the thing that will be billed for (and show up as a line item on an invoice) is `LineItem`

I am making the package service be the source of truth for:

* What to call a line item
* how a line item is derived from usage data

we should not need to define the line item in our protos


proto/sentry_protos/billing/v1/services/package/v1/package.proto
```diff
@@ -0,0 +1,37 @@
...
+  // Base price for the line item (upgraded reserved volumes or add-on activation fees)
+  uint64 base_price_cents = 1;
+  optional uint64 payg_budget_cents = 2;
+  sentry_protos.billing.v1.common.v1.TieredPricingRate payg_rate = 3;
+  sentry_protos.billing.v1.common.v1.TieredPricingRate reserved_rate = 4; // for reserved budget line items
+
+  oneof reserved_units {
+    bool is_unlimited = 5;
+    uint64 num_reserved_units = 6; // the type communicates whether the line item is unlimited or not, additionally reserved budget line items have a non-zero reserved_rate in addition to 0 reserved_units
+  }
```
brendanhsentry: shouldn't this live in the Contract? or will each contract also create a Package?

volokluev: Good call, I think I should change the naming. This should be more like `inlcuded_reserved_units`

proto/sentry_protos/billing/v1/common/v1/billable_metric.proto
```diff
@@ -0,0 +1,57 @@
...
+
+import "sentry_protos/billing/v1/data_category.proto";
+
+// Defines how usage data is transformed into a billable line item through arithmetic operations.
+message BillableMetric {
+  // Unique identifier for the billable metric.
+  string id = 1;
+
+  // Name of the billable metric (e.g., "performance_units").
+  string name = 2;
```
noahsmartin, line 14: What's the purpose of this string? User visible? If it's not user visible do we even need it?

proto/sentry_protos/billing/v1/common/v1/line_item_details.proto
```diff
@@ -0,0 +1,27 @@
...
+import "sentry_protos/billing/v1/common/v1/billable_metric.proto";
+import "sentry_protos/billing/v1/common/v1/unit_info.proto";
+
+// Details of a SKU line item in the billing system.
+message LineItemDetails {
+  // Unique identifier for the line item.
+  string uid = 1;
+
+  // Internal unique name for the SKU.
+  string unique_name = 2;
```
noahsmartin: Same question about this string, what's it for?

proto/sentry_protos/billing/v1/common/v1/unit_info.proto
```diff
@@ -0,0 +1,18 @@
...
+// Unit information for measurement and conversion.
+message UnitInfo {
+  // Short name for the unit (e.g., "GB", "cm", "count").
+  string name = 1;
+
+  // Full descriptive name (e.g., "gigabytes", "centimeters", "count").
+  string full_name = 2;
+
+  // Base unit for conversion (e.g., "byte", "meter", "count").
+  string base_unit = 3;
```
noahsmartin: Can we make this an enum?

volokluev: good call

proto/sentry_protos/billing/v1/common/v1/billable_metric.proto
volokluev: It is user visible, performance units are shown on invoices

proto/sentry_protos/billing/v1/services/package/v1/package.proto
```diff
@@ -0,0 +1,37 @@
...
+  uint64 payg_budget_cents = 2;
+  repeated sentry_protos.billing.v1.common.v1.LineItemDetails line_items = 3;
+}
+
+message PackageConfig {
+  string uid = 1;
+  repeated LineItemConfig line_item_configs = 2;
+  repeated SharedLineItemBudget shared_line_item_budgets = 3;
+  // Base price for the package.
+  uint64 base_price_cents = 4;
```
noahsmartin, line 39: Can we turn base price into a line item? I think it should be representable as an `Expression`. We might have additional kinds of constant non-usage based pricing in the future that would be nice to model all in the same way

proto/sentry_protos/billing/v1/common/v1/billable_metric.proto
```diff
@@ -0,0 +1,57 @@
...
+
+    // A constant numeric value.
+    double constant = 2;
+
+    // A binary arithmetic operation.
+    BinaryOperation binary_op = 3;
+  }
+}
+
+// Reference to a usage data category.
```
noahsmartin, line 34: Is the idea that seat based usage would be added here eventually?

proto/sentry_protos/billing/v1/common/v1/billable_metric.proto
noahsmartin: In that case it probably isn't "performance_units" but "Performance units" right? Maybe we can just call out in the comment that it's user visible. It's also not clear to me how this is differentiated from `customer_facing_name`

py/tests/test_billing_v1.py
```diff
@@ -314,3 +322,69 @@ def test_inactive_credit():
...
+                right=Expression(
+                    category_ref=CategoryReference(
+                        category=DataCategory.DATA_CATEGORY_TRANSACTION_INDEXED
+                    )
+                ),
+            )
+        ),
+    )
+
+    # Verify the metric properties
```
volokluev: verifications are unnecessary, the test is only meant to illustrate how these protos come together

proto/sentry_protos/billing/v1/services/package/v1/package.proto
```diff
@@ -0,0 +1,38 @@
...
+
+// Represents a budget that is collectively used by one or more line items,
+// allowing multiple line items to draw from the same reserved budget.
+message SharedLineItemBudget {
+  uint64 reserved_budget_cents = 1;
+  uint64 payg_budget_cents = 2;
+  repeated sentry_protos.billing.v1.common.v1.LineItemDetails line_items = 3;
+}
+
+message PackageConfig {
```
brendanhsentry, line 34: we'll want a field to specify annual vs monthly packages

proto/sentry_protos/billing/v1/services/package/v1/package.proto
volokluev: On the one hand it makes sense, you define a line item and then it directly becomes an invoice line item, on the other hand, there is a relationship here that I don't think makes sense to discard.

this may be the problem with calling things LineItem(s) instead of SKUs or Products. Because we use this word, we expect that what we define in a package will map 1:1 with an invoice line item (which afaics is a presentation concern), rather than a category of invoice line items. For example: an invoice will have:

```
  metrics (base price): $10
  metrics (reserved_spend): $30
  metrics (payg spend): $20
```

Presented on an invoice, they may all be different lines, but they are part of the same category of line items with a pricing configuration that is defined as one. 

In the package service, I'm not actually describing the line items separately, I describing a product and what its rates are (in this case metrics). 

If I think about driving the checkout page, it makes sense to bundle together the related costs of metrics (base price, included reserved units, reservation costs, and payg costs).

proto/sentry_protos/billing/v1/common/v1/billable_metric.proto
volokluev: good catch, let me add it

proto/sentry_protos/billing/v1/services/package/v1/package.proto
volokluev: I think I also may have mis-perceived what you were commenting on. You meant to make the *package* base price into a lineitem. Let me consider that a bit more. My initial reaction is I'm not opposed to it, and it's also probably not hard to change later

noahsmartin: oh yeah I meant the package base price 👍 (like $29)

proto/sentry_protos/billing/v1/services/package/v1/package.proto
noahsmartin: I envisioned that being on the contract not the package, but thinking about it more I guess the billing cadence determines what the actual price is so maybe it has to be on the package

volokluev: I have added it, good spot

proto/sentry_protos/billing/v1/services/package/v1/package.proto
```diff
@@ -0,0 +1,41 @@
...
+
+package sentry_protos.billing.v1.services.package.v1;
+
+import "sentry_protos/billing/v1/common/v1/billing_interval.proto";
+import "sentry_protos/billing/v1/common/v1/line_item_details.proto";
+import "sentry_protos/billing/v1/common/v1/pricing_tier.proto";
+
+message LineItemConfig {
+  // Base price for the line item (upgraded reserved volumes or add-on activation fees)
+  uint64 base_price_cents = 1;
```
brendanhsentry, line 11: so would we have a LineItemConfig for each reserved tier? ie 50k errors, 100k errors, etc. Could we add a field that specifies the cost for each reserved tier? then just keep `base_price_cents` for add-on activation fees only

proto/sentry_protos/billing/v1/services/package/v1/package.proto
```diff
@@ -0,0 +1,41 @@
...
+    uint64 num_reserved_units = 6; // the type communicates whether the line item is unlimited or not, additionally reserved budget line items have a non-zero reserved_rate in addition to 0 reserved_units
+  }
+
+  sentry_protos.billing.v1.common.v1.LineItemDetails line_item = 7;
+}
+
+// Represents a budget that is collectively used by one or more line items,
+// allowing multiple line items to draw from the same reserved budget.
+message SharedLineItemBudget {
+  uint64 reserved_budget_cents = 1;
```
brendanhsentry: should be reserved_budget_pool; expressed in units not dollars

^ we can do this later

proto/sentry_protos/billing/v1/services/package/v1/package.proto
volokluev: WE talked in person:

reserved tiers are stored in `reserved_rates`, the comment was unhelpful because "reserverd budget line items" has a legacy connotation

proto/sentry_protos/billing/v1/common/v1/billable_metric.proto
volokluev: you're right, removed the two name fields

proto/sentry_protos/billing/v1/common/v1/line_item_details.proto
volokluev: removed

brendanhsentry (approved):

krithikravi (approved):
agree that `BillableMetric.name` and `LineItemDetails.customer_facing_name` are redundant because it's a 1:1 relationship anyway, otherwise lgtm

PR #202, opened 2026-04-18
add proto for requests to new billing platform charge service stripe method

armcknight:
Needed for https://github.com/getsentry/getsentry/pull/19958


volokluev (commented):
I think that this is too specific to stripe. keep in mind that we an make financial transactions through other partner organizations (e.g. vercel). The charge service should more generically encapsulate the "carry through the financial transaction"

PR #205, opened 2026-04-21
feat(billing-platform): Add RateCard protos

brendanhsentry:
closes https://linear.app/getsentry/issue/REVENG-6/define-ratecard-protos

RateCard protos for UsagePricer service. 

It includes reserved amount for each line item, PAYG rates, and shared pools after contract overrides have been applied to the Package.


proto/sentry_protos/billing/v1/services/rate_card/v1/rate_card.proto
```diff
@@ -0,0 +1,35 @@
...
+  sentry_protos.billing.v1.common.v1.TieredPricingRate payg_rate = 4;
+  sentry_protos.billing.v1.common.v1.TieredPricingRate reserved_rate = 5;
+}
+
+message SharedRateCardLineItem {
+  repeated RateCardLineItem rate_card_line_items = 1;
+
+  // Effective reserved_pool_cents after contract overrides have been resolved. If there are no
+  // contract overrides, the default package values are used.
+  oneof reserved_pool_cents {
```
krithikravi, line 26: do we support unlimited reserved budget amounts right now? the package `SharedLineItemPool` does not have an unlimited option for `reserved_pool_cents`

proto/sentry_protos/billing/v1/services/rate_card/v1/rate_card.proto
```diff
@@ -0,0 +1,35 @@
+syntax = "proto3";
+
+package sentry_protos.billing.v1.services.rate_card.v1;
+
+import "sentry_protos/billing/v1/common/v1/line_item_details.proto";
+import "sentry_protos/billing/v1/common/v1/pricing_tier.proto";
+
+message RateCardLineItem {
```
krithikravi, line 8: did we get any clarity on how the hot vs slow path for invoicing will differ? the rate card doesn't have a field for the plan's base price, so we would need to find an alternate method to pass on this information for the slow path

proto/sentry_protos/billing/v1/services/rate_card/v1/rate_card.proto
brendanhsentry: Yeah we don't allow unlimited reserved budgets. But we're planning to use shared line items for other products like performance units which do allow unlimited volume.

proto/sentry_protos/billing/v1/services/rate_card/v1/rate_card.proto
brendanhsentry: Ah yeah I forgot to talk to Volo about this. I was planning to keep the RateCard specific to information we need in the UsagePricer (reserved units, payg rates, etc.) and in the Invoice service fetch the contract and package information to calculate the base prices. The alternative would be to pass base prices along through the RateCard -> UsagePricer -> Invoicer.

@volokluev thoughts?

volokluev: I think that either decision can work and it's not a hard one to change later, just pick something and go with it

proto/sentry_protos/billing/v1/services/rate_card/v1/rate_card.proto
```diff
@@ -0,0 +1,35 @@
...
+  sentry_protos.billing.v1.common.v1.LineItemDetails line_item_detail = 1;
+
+  // The following values are the effective values _after_ contract overrides have been resolved. If there are no
+  // contract overrides, the default package values are used.
+  oneof reserved_units {
+    bool is_unlimited = 2;
+    uint64 num_reserved_units = 3;
+  }
+  sentry_protos.billing.v1.common.v1.TieredPricingRate payg_rate = 4;
```
dashed, line 17: We have products that do not support PAYG such as size analysis. Would this rate card support the rate cards for those products?

krithikravi: hmm i think just leaving the field unset should convey the message that the product does not support payg

proto/sentry_protos/billing/v1/services/rate_card/v1/rate_card.proto
krithikravi: ah in that case we will probably need to update the package proto's `SharedLineItemPool` to support that, i'll leave that change to after V0

krithikravi (approved):

proto/sentry_protos/billing/v1/services/rate_card/v1/rate_card.proto
noahsmartin: In that case it should be optional right?

brendanhsentry: yes krithik is right.

brendanhsentry: @noahsmartin ah yeah let me add that

brendanhsentry: nvm just remembered proto fields are optional by default so we should be all right leaving it as is

volokluev (approved):

PR #206, opened 2026-04-21
feat(billing): Add charge service and capture charge endpoint

noahsmartin:
Introduces a new charge service under billing/v1/services with a CaptureCharge endpoint. The request carries payment intent details and an optional invoice reference; the response reports whether the charge was paid and an optional failure code for downstream error handling.

Two other endpoints that will belong in charge service but we don't need yet for the first milestone are:
- list charges (for a frontend page showing a user their charges)
- charge for invoice_id


proto/sentry_protos/billing/v1/services/charge/v1/endpoint_capture_charge.proto
```diff
@@ -0,0 +1,20 @@
+syntax = "proto3";
+
+package sentry_protos.billing.v1.services.charge.v1;
+
+import "google/protobuf/timestamp.proto";
+
+message CaptureChargeRequest {
+  bool perform = 1;
```
volokluev: what is this for?

proto/sentry_protos/billing/v1/services/charge/v1/endpoint_capture_charge.proto
```diff
@@ -0,0 +1,20 @@
...
+  int64 amount = 2;
```
volokluev: we should specify the unit when we're talking about money. I'm guessing `amount_cents`?

proto/sentry_protos/billing/v1/services/charge/v1/endpoint_capture_charge.proto
```diff
@@ -0,0 +1,20 @@
...
+  string description = 3;
+  string payment_intent = 4;
```
volokluev: what is this? can it be an enum?

proto/sentry_protos/billing/v1/services/charge/v1/endpoint_capture_charge.proto
```diff
@@ -0,0 +1,20 @@
...
+  string organization_id = 5;
```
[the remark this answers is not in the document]
volokluev: seems worth addressing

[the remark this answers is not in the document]
noahsmartin: yep

proto/sentry_protos/billing/v1/services/charge/v1/endpoint_capture_charge.proto
noahsmartin: I'll add some documentation + rework the name. TLDR: it indicates if the stripe charge should be made in addition to the DB model created

proto/sentry_protos/billing/v1/services/charge/v1/endpoint_capture_charge.proto
noahsmartin: yeah good call

proto/sentry_protos/billing/v1/services/charge/v1/endpoint_capture_charge.proto
noahsmartin: it comes from stripe, so I don't think it can, will double check

noahsmartin: Decided to punt on this for now and just removed it. It's definitely not an enum, it's an ID provided by stripe when a payment needs to be retried due to some security verification. Figured that might warrant a separate charge service endpoint but no need to worry about it in phase 1

proto/sentry_protos/billing/v1/services/charge/v1/endpoint_capture_charge.proto
noahsmartin: cleaned this up to an enum, only two values for now but I supposed we could add other payment fulfillment other than stripe in the future

proto/sentry_protos/billing/v1/services/charge/v1/endpoint_capture_charge.proto
```diff
@@ -0,0 +1,29 @@
...
+  // external payment provider. Payment is expected to be collected
+  // externally.
+  CHARGE_METHOD_NONE = 0;
+  // Call the Stripe API to bill the customer for this charge.
+  CHARGE_METHOD_STRIPE = 1;
+}
+
+message CaptureChargeRequest {
+  ChargeMethod charge_method = 1;
+  int64 amount_cents = 2;
```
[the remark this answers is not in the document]
volokluev: Is the idea behind making it an int that refunds are just negative charges?

proto/sentry_protos/billing/v1/services/charge/v1/endpoint_capture_charge.proto
```diff
@@ -0,0 +1,29 @@
...
+  optional string description = 3;
+  uint64 organization_id = 5;
+  google.protobuf.Timestamp current_ts = 6;
+  optional string invoice_id = 7;
+}
+
+message CaptureChargeResponse {
+  bool paid = 1;
+  optional string failure_code = 2;
+}
```
volokluev, line 33: wouldn't you want to return the charge id here?

noahsmartin: I'm hoping it would not be necessary, there isn't any other API that allows you to provide a charge id so not sure why that would need to be exposed

krithikravi (approved):

brendanhsentry (approved):

volokluev (approved):

PR #208, opened 2026-04-21
feat(billing): Define request and response messages for package service

krithikravi:
https://linear.app/getsentry/issue/REVENG-2/create-shim-implementation-of-package-service

Creates request and response message protos for the package service, so we can import them for the package service in getsentry.


proto/sentry_protos/billing/v1/services/package/v1/endpoint_get_package.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto3";
+
+package sentry_protos.billing.v1.services.package.v1;
+
+import "sentry_protos/billing/v1/services/package/v1/package.proto";
+
+message GetPackageRequest {
+  uint64 package_id = 1;
```
brendanhsentry: believe this should be a string and called `package_uid`
https://github.com/getsentry/sentry-protos/blob/5b5799c73d60496f957dc6e821532eca4dceee18/proto/sentry_protos/billing/v1/services/package/v1/package.proto#L34-L41

brendanhsentry (approved):
lgtm

volokluev (approved):

PR #211, opened 2026-04-22
feat(billing): Add invoice, list charges, and rollover endpoints

noahsmartin:
Add new endpoints and supporting messages for the charge and contract services:

- charge: ListChargesForInvoice
This is used by the invoicer when we get uncharged invoices. We first list all charges to confirm that it is really not charged yet (in case the charge was written but we failed to update the invoice model)

- contract: GetInvoice, GetUnchargedInvoices, RolloverContract
get invoice is used by the invoicer in an async task that will be provided just the invoice_id
get uncharged invoices is the main mechanism for charging (and also retrying) similar to how we get contracts that aren't invoiced
rollover contract is the API for creating a new billing period and creating the invoice model (done as one operation to avoid race conditions/satisfy FK requirements)


proto/sentry_protos/billing/v1/services/charge/v1/endpoint_list_charges.proto
```diff
@@ -0,0 +1,19 @@
...
+message Charge {
+    uint64 amount_cents = 1;
+    bool paid = 2;
+    optional string failure_code = 3;
+}
+
+// No pagination on this request because there are expected to be only a handful
+// of attempted charges per invoice.
+message ListChargesForInvoiceRequest {
+    uint64 invoice_id = 1;
```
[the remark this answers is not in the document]
noahsmartin: The new types are all correct. An old `string` type got through which is unfortunate but can be handled since it's easy to convert an int to a string

proto/sentry_protos/billing/v1/services/contract/v1/endpoint_get_uncharged_invoices.proto
```diff
@@ -0,0 +1,26 @@
...
+
+  optional uint32 offset = 3;
+}
+
+message GetUnchargedInvoicesResponse {
+  repeated uint64 invoice_ids = 1;
+
+  // True if additional matching invoices existed beyond max_items and were
+  // not included in this response.
+  bool truncated = 2;
```
volokluev, line 25: why not just return what the next offset should be

proto/sentry_protos/billing/v1/services/contract/v1/invoice.proto
```diff
@@ -0,0 +1,18 @@
...
+message InvoiceLineItem {
+    // Intentionally not a uint (some line items could be discounts)
+    int64 amount_cents = 1;
+    optional string description = 2;
+}
+
+message Invoice {
+    uint64 invoice_id = 1;
+    repeated InvoiceLineItem line_items = 2;
+    // Not just a sum of line items since there may be credit applied
```
brendanhsentry, line 16: currently, we create invoice line items for credits which I think makes it easier to understand everything that went into the final amount billed.

volokluev (approved):
overall a good start

brendanhsentry (commented):
why are all the invoice protos in the contract directory?

volokluev:
> why are all the invoice protos in the contract directory?

we're keeping the data in the contract service because the contract is highly tied to the billing period either way. The invoice service just mediates the tallying and charging but doesn't store anything

proto/sentry_protos/billing/v1/services/contract/v1/endpoint_get_uncharged_invoices.proto
noahsmartin: I'm not opposed to it, but the other pagination API already uses this "truncated" boolean. Seems like matching it is better than starting a new standard

proto/sentry_protos/billing/v1/services/contract/v1/invoice.proto
noahsmartin: This could be like an overpayment that gets applied, not a typical "credit". Maybe those will also end up in the line items too though

PR #212, opened 2026-04-23
feat(billing-platform): Update UsagePricer protos

brendanhsentry:
closes https://linear.app/getsentry/issue/REVENG-13/redefine-usagepricer-output-proto

Updates UsagePricer to use line items instead of SKUs

### DIFF
--- a/proto/sentry_protos/billing/v1/services/usage_pricer/v1/endpoint_usage_pricer.proto
+++ b/proto/sentry_protos/billing/v1/services/usage_pricer/v1/endpoint_usage_pricer.proto
@@ -15,6 +15,7 @@ message GetPriceForContractRequest {
   uint64 contract_id = 1;
 }
 
+// DEPRECATED: use LineItemUsageSummary instead
 message SKUUsageSummary {
   sentry_protos.billing.v1.SKU sku = 1;
   // Net cents consumed by this SKU in the billing period (after credits/trials applied).
@@ -23,7 +24,7 @@ message SKUUsageSummary {
   uint64 usage_volume = 3;
 }
 
-// Pricing breakdown for a shared budget spanning multiple SKUs.
+// DEPRECATED: use SharedLineItemUsageSummary instead
 message SharedBudgetUsageSummary {
   repeated sentry_protos.billing.v1.SKU skus = 1;
   // Net cents consumed across all SKUs in this shared budget (after credits/trials applied).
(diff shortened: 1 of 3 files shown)


proto/sentry_protos/billing/v1/services/usage_pricer/v1/endpoint_usage_pricer.proto
```diff
@@ -32,9 +33,28 @@ message SharedBudgetUsageSummary {
...
+}
+
 message UsagePricerResponse {
-  // Per-SKU pricing breakdown.
+  // DEPRECATED: use line_item_summaries
   repeated SKUUsageSummary sku_summaries = 1;
-  // Per-shared-budget pricing breakdown (for SKUs sharing a budget).
+
+  // DEPRECATED: use shared_line_item_summaries
   repeated SharedBudgetUsageSummary shared_budget_summaries = 2;
```
volokluev: there's a deprecated marker in protobufs use that

```
int32 old_field = 6 [deprecated = true];
```

volokluev:
@noahsmartin please review this and see if this gives you all the information you need for invoicing. notice that the results only exposes the payg spend, not the reservation spend (although those are billed at different times). We should make sure we are aligned on service expectations

proto/sentry_protos/billing/v1/services/usage_pricer/v1/endpoint_usage_pricer.proto
brendanhsentry: just adding `option deprecated = true;` should enough deprecate the whole message right?

volokluev: possible, give it a shot

krithikravi:
> @noahsmartin please review this and see if this gives you all the information you need for invoicing. notice that the results only exposes the payg spend, not the reservation spend (although those are billed at different times). We should make sure we are aligned on service expectations

yep, this means we need different invoicing paths for the hot path(payg costs) and the cold path(monthly/reserved costs). we also don't store plan base price info in the rate card iirc

proto/sentry_protos/billing/v1/services/usage_pricer/v1/endpoint_usage_pricer.proto
```diff
@@ -16,25 +16,49 @@ message GetPriceForContractRequest {
...
 
+message LineItemUsageSummary {
+  string line_item_uid = 1;
+  // Net cents consumed by this line item in the billing period (after credits/trials applied).
+  uint64 payg_spend_cents = 2;
+}
+
+message SharedLineItemUsageSummary {
+  // Net cents consumed across all SKUs in this shared budget (after credits/trials applied).
+  uint64 payg_spend_cents = 1;
```
noahsmartin, line 50: Is this not just the sum of all the line_item_summaries `payg_spend_cents` fields? If so I'd recommend omitting it so there isn't duplicated state that would have the potential to get out of sync

proto/sentry_protos/billing/v1/services/usage_pricer/v1/endpoint_usage_pricer.proto
```diff
@@ -16,25 +16,49 @@ message GetPriceForContractRequest {
...
+
   repeated sentry_protos.billing.v1.SKU skus = 1;
   // Net cents consumed across all SKUs in this shared budget (after credits/trials applied).
   uint64 payg_spend_cents = 2;
   // Per-SKU breakdown within the shared budget.
   repeated SKUUsageSummary sku_summaries = 3;
 }
 
+message LineItemUsageSummary {
+  string line_item_uid = 1;
```
noahsmartin, line 42: Is this uid supposed to be used to identify some other object? Like something in the package service? I imagine it will be so that we can correlate the price here to a human readable description of the line item. Can you leave a comment indicating which object it's a pointer to?

noahsmartin (approved):

proto/sentry_protos/billing/v1/services/usage_pricer/v1/endpoint_usage_pricer.proto
brendanhsentry: There could be shared line item level credits

PR #214, opened 2026-04-24
feat: Add billing details service interface

noahsmartin:


proto/sentry_protos/billing/v1/services/contract/v1/billing_config.proto
brendanhsentry, line 44: should we mark this message as deprecated as well

brendanhsentry (approved):

proto/sentry_protos/billing/v1/services/contract/v1/billing_config.proto
noahsmartin: yep, done

PR #216, opened 2026-04-27
ref: Update contract protos

noahsmartin:
Add the new Contract proto that will work with the package. Deprecated old fields that are now part of package

https://linear.app/getsentry/issue/REVENG-17/redefine-contract-protos

### DIFF
--- a/proto/sentry_protos/billing/v1/services/contract/v1/pricing_config.proto
+++ b/proto/sentry_protos/billing/v1/services/contract/v1/pricing_config.proto
@@ -2,21 +2,29 @@ syntax = "proto3";
 
 package sentry_protos.billing.v1.services.contract.v1;
 
+import "google/protobuf/empty.proto";
+import "google/protobuf/timestamp.proto";
 import "sentry_protos/billing/v1/services/contract/v1/billing_config.proto";
 import "sentry_protos/billing/v1/services/contract/v1/sku.proto";
 import "sentry_protos/billing/v1/sku.proto";
 
 message PricingTier {
+  option deprecated = true;
+
   int64 start = 1;
   int64 end = 2;
   int64 rate_per_unit_cpe = 3;
 }
 
 message TieredPricingRate {
+  option deprecated = true;
+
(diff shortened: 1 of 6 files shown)


proto/sentry_protos/billing/v1/services/contract/v1/pricing_config.proto
```diff
@@ -36,24 +43,47 @@ message SKUConfig {
...
+  repeated uint64 billing_metrics = 2;
+}
+
+message Reservation {
+  uint64 reserved_price_cents = 1;
+  oneof reserved_units {
+    bool is_unlimited = 2;
+    uint64 num_reserved_units = 3;
+  }
+  repeated uint64 billing_metrics = 4;
```
brendanhsentry: is `billing_metrics` a list of proto DataCategories? Also will this list ever differ from `PAYGBudget.billing_metrics`? if not, we can move this to the `UserConfig` message

edit: just realized we'd need to make UserConfig repeated in PricingConfig and make PAYGBudget and Reservation not repeated in UserConfig for what i said to work

proto/sentry_protos/billing/v1/services/contract/v1/pricing_config.proto
```diff
@@ -36,24 +43,47 @@ message SKUConfig {
...
+}
+
+message UserConfig {
+  repeated PAYGBudget payg_budget = 1;
+  repeated Reservation reservation = 2;
+}
```
brendanhsentry, line 78: Can we include a `string line_item_uid` field in UserConfig? It would make a lot easier to look up overrides in the RateCard service.

proto/sentry_protos/billing/v1/services/contract/v1/pricing_config.proto
noahsmartin: This should have been line_item_uid, updated

brendanhsentry (approved):
protos lgtm. cursor's comments about the test are legit though

PR #219, opened 2026-04-27
ref: Add plan family to package

noahsmartin:


proto/sentry_protos/billing/v1/services/package/v1/package.proto
```diff
@@ -31,6 +31,14 @@ message SharedLineItemPool {
...
 }
 
+enum PlanFamily {
+  PLAN_FAMILY_UNSPECIFIED = 0;
+  PLAN_FAMILY_DEVELOPER = 1;
+  PLAN_FAMILY_TEAM = 2;
+  PLAN_FAMILY_BUSINESS = 3;
+  PLAN_FAMILY_ENTERPRISE = 4;
+}
+
```
volokluev: I don't think this enum should actually be defined on the proto interface. What's the reasoning behind this?

noahsmartin: For the user facing elements that reflect the plan family. Like the title of the subscription page (we aren't building that yet) or the line item on the invoice (that's why I came across this). I also considered adding something like `string plan_title` but that started to feel too much like server driven UI which tends to be over engineered. I could imagine many copy variants all driven by this family and didn't want to require updating the proto definition every time we found a new UI element driven by this. What do you think? Any other preferred way of handling it?

volokluev: I would much rather put `string plan_title` on the package and let the package service decide how to deal with it rather than the proto. It's why I moved SKU definitions out of the protos and into the package service so we wouldn't have to update something in two places. 

Also at the moment I don't think we're worrying about the checkout page so let's cross that bridge when we get there?

noahsmartin: It shows on the invoice though, something like "Subscription to Team". I'm fine with adding it as a `plan_title`

noahsmartin: updated!

volokluev (approved):

PR #221, opened 2026-04-28
feat: Add create contract endpoint

noahsmartin:


proto/sentry_protos/billing/v1/services/contract/v1/endpoint_create_contract.proto
```diff
@@ -0,0 +1,17 @@
...
+
+package sentry_protos.billing.v1.services.contract.v1;
+
+import "sentry_protos/billing/v1/common/v1/billing_interval.proto";
+import "sentry_protos/billing/v1/services/contract/v1/pricing_config.proto";
+
+message CreateContractRequest {
+    uint64 organization_id = 1;
+    string package_uid = 2;
+    sentry_protos.billing.v1.common.v1.BillingInterval interval = 3;
```
brendanhsentry: the Package will also have a `BillingInterval` so we should be able to omit `interval`

proto/sentry_protos/billing/v1/services/contract/v1/endpoint_create_contract.proto
```diff
@@ -0,0 +1,17 @@
...
+    repeated UserConfig user_config = 4;
```
brendanhsentry:
```suggestion
    repeated UserConfig user_configs = 4;
```

brendanhsentry (approved):
lgtm; minor comments

proto/sentry_protos/billing/v1/services/contract/v1/endpoint_create_contract.proto
noahsmartin: yeah good catch

PR #226, opened 2026-04-28
feat(snuba): add RequestMeta.standard_retention_days

MeredithAnya:
## Summary
Adds an optional `standard_retention_days` field on `RequestMeta`.

Callers should send the organization's standard retention window on each request so Snuba can make better storage routing decisions. With that value, Snuba knows how long full-fidelity data is expected to be available for the org and can avoid downsampling queries that still fall inside that window.

When set, Snuba may keep queries within this window on tier 1 storage instead of applying long-term retention downsampling. When unset, Snuba falls back to its default lower retention (typically 30 days).

## Why this name / placement
- Kept on `RequestMeta`, **not** inside `DownsampledStorageConfig`
- `DownsampledStorageConfig` controls sampling/routing **mode**
- This field is a separate policy input: the org's standard retention window, used as the age threshold for full-fidelity data
- Named to avoid confusion with ingest/TTL `retention_days` on `TraceItem`

## Shape
```protobuf
optional uint32 standard_retention_days = 13;
```

`optional` so consumers can distinguish unset from an explicit value via presence checks (`HasField` / `Option`).

## Consumers
- Snuba: [getsentry/snuba#7912](https://github.com/getsentry/snuba/pull/7912)
- Callers (e.g. Sentry) should pass the organization's standard retention so routing can honor longer full-fidelity windows when available

## Notes
- Field number `13` on `RequestMeta`
- Non-breaking proto3 additive change
- Rust bindings updated to match


phacops:
Note on the earlier Bugbot comment: the field was already made `optional uint32` in a follow-up commit, so unset vs explicit value is distinguishable via presence checks.

phacops (approved):

PR #228, opened 2026-04-29
ref: Use display units

noahsmartin:
If num_reserved_units was meant to be in display units not base units then we need to represent it as a float (ie 5.1 GB)


noahsmartin:
We decided to keep this field as base units and not have display units in the protos https://www.notion.so/sentry/When-to-use-converted-units-3518b10e4b5d80d0867adbb32fbdf209?v=95f8b10e4b5d8290b3b388bec3fc369a&source=copy_link

PR #229, opened 2026-04-29
feat(billing): Add payment config to charge service

volokluev:
### Human comments

First pass of implementing [charging](https://github.com/getsentry/getsentry/pull/20140), I realized that the invoicer needs to pass stripe details to the charge service as well as the invoice guid. This is to avoid having to call the contract service from the charge service. 

### AI comments

Add `PaymentConfig` message to the charge service interface to support Stripe payment details when capturing charges.

The new `payment_config.proto` introduces:
- `StripePaymentData`: Contains customer Stripe ID, default payment method ID, and payment method status
- `PaymentConfig`: Uses a oneof pattern for the payment provider config, allowing future extension to other providers (PayPal, Braintree, etc.)

`CaptureChargeRequest` now accepts an optional `payment_config` field, which services can use to pass provider-specific payment information when `charge_method` is set to `CHARGE_METHOD_STRIPE`.

This design keeps the interface extensible without breaking existing callers who don't need payment provider details.


proto/sentry_protos/billing/v1/services/charge/v1/endpoint_capture_charge.proto
```diff
@@ -25,6 +26,8 @@ message CaptureChargeRequest {
   uint64 organization_id = 4;
   google.protobuf.Timestamp current_ts = 5;
   optional string invoice_id = 6;
+  optional string invoice_guid = 7;
```
noahsmartin, line 29: Ah yes because this goes to stripe for the webhooks 👍

noahsmartin (approved):

proto/sentry_protos/billing/v1/common/v1/payment_config.proto
```diff
@@ -0,0 +1,18 @@
+syntax = "proto3";
+
+package sentry_protos.billing.v1.common.v1;
+
+// Stripe-specific payment information for an organization.
+message StripePaymentData {
+  optional string customer_stripe_id = 1;
+  optional string default_payment_method_id = 2;
+  bool has_payment_method = 3;
```
noahsmartin: What's the difference between `has_payment_method` and `default_payment_method_id` being non-null? Can you ever had `default_payment_method_id = None` and `has_payment_method = True`?

volokluev: you're right it's redundant as of this moment. good catch

PR #236, opened 2026-05-01
ref: Add address fields to invoice

noahsmartin:
The invoice stores the address that was used for billing at the time the invoice was created. This lets us pass it to the contract service so it can be stored on the Invoice model and retrieved from the contract service when rendering the invoice PDF


volokluev (approved):
one for the road

brendanhsentry (approved):

PR #250, opened 2026-05-08
feat(explore): Add optional index field to AttributeKey for array element access

manessaraj:
Adds `optional sint32 index` on `AttributeKey` to support indexed access into array-typed attributes (e.g. `field[N]:value` syntax).


onewland (changes requested):
I don't think we should support negative indexing. I'm actually curious what the use-case is for this overall

PR #253, opened 2026-05-11
feat(billing): added QuotaConfig protos(REVENG-74)

krithikravi:
https://linear.app/getsentry/issue/REVENG-74/create-quotaconfig-protos

This PR adds QuotaConfig protos and tests to validate it. QuotaConfigs will be used to represent rate limiting decisions due to the account status being disabled, payg being disabled, and the user going over quota, among other reasons. This proto closely follows the existing QuotaConfig object from https://github.com/getsentry/sentry/blob/52e4ff5f57db1d13508f4927641f458c45594fcf/src/sentry/quotas/base.py#L123.


proto/sentry_protos/billing/v1/quota_config.proto
```diff
@@ -0,0 +1,92 @@
...
+  // Project-level quota usage exceeded.
+  REASON_CODE_PROJECT_QUOTA_ERROR_USAGE_EXCEEDED = 32;
+  REASON_CODE_PROJECT_QUOTA_ATTACHMENT_USAGE_EXCEEDED = 33;
+  REASON_CODE_PROJECT_QUOTA_TRANSACTION_USAGE_EXCEEDED = 34;
+  REASON_CODE_PROJECT_QUOTA_REPLAY_USAGE_EXCEEDED = 35;
+  REASON_CODE_PROJECT_QUOTA_PROFILE_USAGE_EXCEEDED = 36;
+  REASON_CODE_PROJECT_QUOTA_PROFILE_DURATION_USAGE_EXCEEDED = 37;
+  REASON_CODE_PROJECT_QUOTA_PROFILE_DURATION_UI_USAGE_EXCEEDED = 38;
+  REASON_CODE_PROJECT_QUOTA_SPAN_USAGE_EXCEEDED = 39;
+  REASON_CODE_PROJECT_QUOTA_LOG_BYTE_USAGE_EXCEEDED = 40;
```
noahsmartin: Since one of our goals is to launch a sku without a code change I wonder if we can get rid of these hardcoded reasons? Because we’d want to be able to launch a new sku without adding to this

krithikravi: I considered making the reason code field a string instead of an enum type, but it looks like we have some Datadog metrics that track this field. If we want to remove the hardcoded strings, we will lose information by replacing them with a generic `CATEGORY_USAGE_EXCEEDED`, or we will need to use strings and just be careful with the way we format them when we populate this field.

proto/sentry_protos/billing/v1/quota_config.proto
```diff
@@ -0,0 +1,42 @@
...
+//
+// Three limit states:
+//   - absent:   unlimited (counted but never rejected)
+//   - 0:        reject-all (statically enforced, no window needed)
+//   - positive: enforcement threshold within the given window
+message QuotaConfig {
+  string id = 1;
+  repeated DataCategory categories = 2;
+  QuotaScope scope = 3;
+  optional string scope_id = 4;
```
noahsmartin, line 28: What’s this for? Redundant with with the scope enum?

noahsmartin (approved):

proto/sentry_protos/billing/v1/quota_config.proto
krithikravi: If the scope was a project, then this would reference the project_id, same for an organization etc. I believe relay uses these to actually carry out the rate limiting, but not super familiar with that part of our codebase.

PR #254, opened 2026-05-11
feat(billing): added account protos(REVENG-79)

krithikravi:
https://linear.app/getsentry/issue/REVENG-79/create-accountstatus-proto

This PR adds protos for an account's status, which represent org-wide information not contained on the contract such as org status, soft cap eligibility etc.


proto/sentry_protos/billing/v1/services/account_status/v1/account_status.proto
```diff
@@ -0,0 +1,31 @@
...
+enum OnDemandStatus {
+  ON_DEMAND_STATUS_UNSPECIFIED = 0;
+  ON_DEMAND_STATUS_ENABLED = 1;
+  ON_DEMAND_STATUS_DISABLED = 2;
+}
+
+message AccountStatus {
+  // The organization this status belongs to. Each org has one AccountStatus.
+  uint64 organization_id = 1;
+  // Legacy soft cap flag. When true, no category is rate-limited regardless of usage.
```
noahsmartin: Do we still need a legacy flag on billing platform?

proto/sentry_protos/billing/v1/services/account_status/v1/account_status.proto
```diff
@@ -0,0 +1,31 @@
...
+  bool soft_cap = 2;
+  // Current account status (active, closed, past-due).
+  Status status = 3;
+  // Whether the account is suspended by an admin (e.g., for fraud).
+  bool is_suspended = 4;
```
noahsmartin: Is this redundant with the status enum?

noahsmartin (approved):

proto/sentry_protos/billing/v1/services/account_status/v1/account_status.proto
krithikravi: we currently store this in the existing codebase separately, but i'm fine with merging them tbh

proto/sentry_protos/billing/v1/services/account_status/v1/account_status.proto
krithikravi: This is just referencing the fact that softcaps can now be enabled per category, previously the 'legacy' version applied to the whole subscription and just meant we didn't rate limit them at all. We still use this for use cases like giving an unlimited trial to enterprise orgs that are on a subscription trial etc.

noahsmartin: I think that will end up being managed by the new trials service and not by this flag now

PR #260, opened 2026-05-14
ci(release): Auto-release when proto schemas change on main

armenzg:
Automate releases on merges to `main` when the merge includes protobuf schema changes under `proto/**`, so version bumps and package publishing stay aligned with schema updates.

**Scoped auto-release trigger**

The `release` workflow now runs on `push` to `main` with a `paths` filter for `proto/**`, and uses `version=auto` by default for push-triggered runs.

**Manual override and loop prevention**

Manual `workflow_dispatch` remains available for explicit `version`/`force` inputs, and release-generated `release:` commits are ignored to prevent recursive runs.

README now documents the new trigger scope and the manual dispatch path.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,6 +1,9 @@
 name: release
 
 on:
+  push:
+    branches:
+      - main
   workflow_dispatch:
     inputs:
       version:
@@ -18,6 +21,9 @@ jobs:
   release:
     runs-on: ubuntu-latest
     name: 'Release a new version'
+    if: |
+      github.event_name == 'workflow_dispatch' ||
+      !startsWith(github.event.head_commit.message, 'release:')
     steps:
       - name: Get auth token
         id: token
@@ -36,5 +42,5 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ steps.token.outputs.token }}
         with:
-          version: ${{ github.event.inputs.version }}
-          force: ${{ github.event.inputs.force }}
+          version: ${{ github.event.inputs.version || 'auto' }}
+          force: ${{ github.event.inputs.force || 'false' }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - main
+    paths:
+      - 'proto/**'
   workflow_dispatch:
     inputs:
       version:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -1,6 +1,9 @@
 name: release
 
 on:
+  push:
+    branches:
+      - main
```
armenzg, line 6: Execute on pushes on `main` besides when dispatched via the workflow.

.github/workflows/release.yml
```diff
@@ -1,6 +1,11 @@
...
+    paths:
+      - 'proto/**'
```
armenzg, line 8: We will only do releases if changes happen under `proto/`

.github/workflows/release.yml
```diff
@@ -18,6 +23,9 @@ jobs:
   release:
     runs-on: ubuntu-latest
     name: 'Release a new version'
+    if: |
+      github.event_name == 'workflow_dispatch' ||
+      !startsWith(github.event.head_commit.message, 'release:')
```
armenzg, line 28: These lines are a **job-level guard** that decides whether the `release` job should run:

- `github.event_name == 'workflow_dispatch'`  
  - If someone manually triggers the workflow from Actions UI, always run.

- `||`  
  - Logical OR.

- `!startsWith(github.event.head_commit.message, 'release:')`  
  - For push-triggered runs, skip the job when the latest commit message starts with `release:`.

Why this exists:

- Your automated release flow creates commits like `release: x.y.z`.
- Those commits also land on `main` and would retrigger the workflow.
- This check prevents that self-triggering loop.

So in practice: **manual runs always execute; push runs execute unless they are release-generated commits.**

.github/workflows/release.yml
```diff
@@ -36,5 +44,5 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ steps.token.outputs.token }}
         with:
-          version: ${{ github.event.inputs.version }}
-          force: ${{ github.event.inputs.force }}
+          version: ${{ github.event.inputs.version || 'auto' }}
+          force: ${{ github.event.inputs.force || 'false' }}
```
armenzg, line 48: Default to these values if not defined.

noahsmartin (approved):
Thank you!

.github/workflows/release.yml
```diff
@@ -1,6 +1,11 @@
 name: release
 
 on:
+  push:
+    branches:
+      - main
+    paths:
```
armenzg, line 7: Hello world!

PR #265, opened 2026-05-25
feat(billing-platform): Add billing_interval snapshot to Contract BillingConfig

armenzg:
Adds a `BillingInterval billing_interval` field (number 7) to `BillingConfig` so a contract can carry the cadence it was signed under on the wire. The contract domain in getsentry will populate this from a snapshot taken at signup, so the cadence becomes a frozen attribute of the contract instead of something callers have to re-resolve from the package on every read.

**Scope**

Single-file proto change. The `BillingInterval` enum already exists in `proto/sentry_protos/billing/v1/common/v1/billing_interval.proto` (`UNSPECIFIED=0`, `MONTHLY=1`, `ANNUAL_BASE_MONTHLY_PAYG=2`); this PR just imports it and adds the field with a doc comment that it's frozen for the life of the contract.

**Consumer**

This is preparatory infrastructure for the consumer in getsentry/getsentry#20403, which itself is prep work for the larger annual-contract feature in getsentry/getsentry#20333. After this PR releases, #20403 will repin sentry-protos and start populating `Contract.billing_config.billing_interval`.

Refs getsentry/getsentry#20403
Refs getsentry/getsentry#20333

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -717,7 +717,7 @@ checksum = "955d28af4278de8121b7ebeb796b6a45735dc01436d898801014aced2773a3d6"
 
 [[package]]
 name = "sentry_protos"
-version = "0.11.0"
+version = "0.13.0"
 dependencies = [
  "prost",
  "prost-types",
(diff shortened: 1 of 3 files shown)


armenzg:
We have a new plan.

PR #267, opened 2026-05-26
feat(billing-platform): add charge-disputed handler + charge lookup endpoints

armcknight:
Adds two proto endpoints needed for the platform-side charge.dispute.created webhook handler:

- InvoicerService.HandleChargeDisputed: lets the legacy webhook ask the billing platform whether a disputed Stripe charge belongs to the platform and, if so, hand suspension off to the platform.
- ChargeService.GetChargeByStripeId: read-only lookup so the dispute handler can identify platform charges without mutating their stored state (unlike UpdateCharge, which expects a full StripeCharge payload). 
 
REVENG-89


armcknight:
I'm unsure we need the ChargeService.GetChargeByStripeId, but maybe i just don't understand the implications of the current API

noahsmartin (approved):

PR #268, opened 2026-05-26
feat(billing): Add GrantSource, source field, and GetEffectiveGrants protos

dashed:
## Summary

Adds proto definitions for the Engagement Service (Approach 2 scope).

### Changes

| File | Change | Purpose |
|------|--------|---------|
| `grant_source.proto` | New file | `GrantSource` enum (7 values: trial, promo, recurring, gift) |
| `grant.proto` | Updated | Added `GrantSource source = 9` field + import |
| `endpoint_get_effective_grants.proto` | New file | `EffectiveGrant`, `GetEffectiveGrantsRequest/Response` — returns grants with counter-chains collapsed for caller-side drain |

### Dependencies

- `grant.proto` — already merged (Grant, GrantType, GrantStatus messages); updated here to add source field
- `date.proto` — already in the repo (shared Date type)

## Test plan

- [x] buf lint passes
- [x] buf format passes
- [ ] `make build-py` compiles all protos


proto/sentry_protos/billing/v1/services/engagement/v1/grant.proto
```diff
@@ -35,4 +36,5 @@ message Grant {
   sentry_protos.billing.v1.Date start_date = 6;
   sentry_protos.billing.v1.Date end_date = 7;
   GrantStatus status = 8;
+  GrantSource source = 9;
```
dashed, line 48: this isnt needed for usage pricer. but adding this here because source field is in the Grant table that's being introduced in https://github.com/getsentry/getsentry/pull/20409. just minimizing the amount of potential database migrations we would need to do later in the future.

proto/sentry_protos/billing/v1/services/engagement/v1/endpoint_get_effective_grants.proto
```diff
@@ -0,0 +1,28 @@
...
+  // May differ from grant.amount when counter-grants exist.
+  int64 effective_amount = 2;
+}
+
+message GetEffectiveGrantsRequest {
+  uint64 organization_id = 1;
+  // When set, returns only grants whose line_item_uids contain this value.
+  optional string line_item_uid = 2;
+  sentry_protos.billing.v1.Date start_date = 3;
+  sentry_protos.billing.v1.Date end_date = 4;
```
brendanhsentry, line 35: will this be inclusive? for example, what if a grant (ie product trial) starts before a billing period and ends in the middle of it? the UsagePricer service will most likely request start_date equal to the contact's period start, but we should expect that product trial to be included.

We don't need to change the structure of this message, but something to keep in mind. Maybe a comment would be useful.

proto/sentry_protos/billing/v1/services/engagement/v1/endpoint_get_effective_grants.proto
```diff
@@ -0,0 +1,28 @@
...
+import "sentry_protos/billing/v1/date.proto";
+import "sentry_protos/billing/v1/services/engagement/v1/grant.proto";
+
+// A grant with counter-grant chains collapsed into a single effective amount.
+// Returned in drain priority order (end_date ASC, effective_amount ASC, id ASC).
+message EffectiveGrant {
+  Grant grant = 1;
+  // The usable amount after counter-grant chain collapse.
+  // May differ from grant.amount when counter-grants exist.
+  int64 effective_amount = 2;
```
brendanhsentry: which grant is returned here if there's a counter-grant? or will there be multiple EffectiveGrants with the same effective_amount?

Also, I'm not sure we need to represent that in the protobuf message. We could use this append-only log system you've described in our actual db records, but keep this particular protobuf simple so it only return what we care about for the UsagePricer. If we need to output each Grant and counter-grant in the future, we can add a new endpoint and message to represent that.

proto/sentry_protos/billing/v1/services/engagement/v1/endpoint_get_effective_grants.proto
dashed: yes these dates are inclusive.

we return effective grants (e.g. product trial grant) if their `[start_date, end_date]` overlaps with the requested `[requested_start_date, requested_end_date]` window. essentially,  `grant.start_date <= request.end_date`  and `grant.end_date >= request.start_date`.

something to note that `EffectiveGrant`'s start and end dates are not trimmed to the start/end dates of `GetEffectiveGrantsRequest`.

proto/sentry_protos/billing/v1/services/engagement/v1/endpoint_get_effective_grants.proto
dashed: > which grant is returned here if there's a counter-grant?

it would be the "root" grant.

Grants are effectively immutable; so the only way to end a trial early is to create a counter grant to cancel out the remaining credits of the original grant.

I've originally had a `parent_id` in the `Grant` proto for these use cases. But I've slimmed it down to what the usage pricer needs right now. However the parent id field exists in the Grant table in https://github.com/getsentry/getsentry/pull/20409.

I aim to introduce more fields to the `Grant` proto model in later PRs.

for the purpose of EffectiveGrant, we fold all the counter grants together (i.e. summed them up) against the primary grant to get the `effective_amount`.

this lets the caller to consume one grant, rather than a grant + any of their counter grants.

------

we do this simplification vs the other approach where you just input daily usage, and get "credited" daily usage back. this was the original approach i had in https://github.com/getsentry/getsentry/pull/20410.

brendanhsentry (approved):

PR #269, opened 2026-05-27
ref(billing): split package support and contract month interval

armenzg:
Packages now advertise the set of cadences they offer (`supported_month_intervals`) instead of a single enum interval, and contract creation snapshots the chosen cadence on the contract so it stays stable for the contract's lifetime even if the package later changes its offerings.

**Package cadence is now a list**

`PackageConfig.billing_interval` (and the `BillingInterval` enum) are deprecated. Callers should read `supported_month_intervals`, whose first element is treated as the package default.

Before:

```proto
message PackageConfig {
  sentry_protos.billing.v1.common.v1.BillingInterval billing_interval = 5;
  // ...
}
```

After:

```proto
message PackageConfig {
  sentry_protos.billing.v1.common.v1.BillingInterval billing_interval = 5 [deprecated = true];
  repeated uint32 supported_month_intervals = 7; // e.g. [1], [12], [1, 12]
  // ...
}
```

**Contract cadence is selected at create time and frozen**

`CreateContractRequest.month_interval` lets the caller pick one of the package's supported cadences. The selection is recorded on `BillingConfig.month_interval` and is not expected to change for the life of the contract.

```proto
message CreateContractRequest {
  uint32 month_interval = 6; // must be one of PackageConfig.supported_month_intervals
}

message BillingConfig {
  uint32 month_interval = 7; // frozen for the contract's lifetime
}
```


proto/sentry_protos/billing/v1/common/v1/billing_interval.proto
armenzg, line 1: Once we migrate to the new version we will remove this completely.

proto/sentry_protos/billing/v1/services/contract/v1/billing_config.proto
```diff
@@ -57,4 +57,8 @@ message BillingConfig {
   // Use PricingConfig.billing_period_start_date and PricingConfig.billing_period_end_date
   Date contract_start_date = 5 [deprecated = true];
   Date contract_end_date = 6 [deprecated = true];
+
+  // The number-of-months interval the contract was signed under
+  // (1 = monthly, 12 = annual). Frozen for the life of the contract.
+  uint32 month_interval = 7;
```
armenzg, line 63: This will allow us to support any interval we want including multi-year contracts.

krithikravi (approved):

brendanhsentry (approved):

PR #270, opened 2026-05-27
feat(devenv): guard sentry-protos worktree basename mismatches

armenzg:
Prevent `direnv` from failing with a Python `KeyError` when a `sentry-protos` worktree is created under a directory whose basename does not match the expected `venv` section name.

When the basename check fails, `.envrc` now prints a clear remediation message and skips `devenv sync` instead of attempting a broken sync path.

**Safer Worktree Startup**

This makes misnamed worktrees fail soft with actionable guidance, while preserving the existing setup flow for correctly named checkouts.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -2,7 +2,12 @@ if [ -f .env ]; then
     dotenv
 fi
 
-if ! command -v "devenv" >/dev/null; then
+expected_repo_name="$(awk -F'[][.]' '/^\[venv\./{print $3; exit}' devenv/config.ini 2>/dev/null)"
+
+if [ -n "$expected_repo_name" ] && [ "$(basename "$PWD")" != "$expected_repo_name" ]; then
+  echo "direnv: worktree basename '$(basename "$PWD")' does not match expected '$expected_repo_name'."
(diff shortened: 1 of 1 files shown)


.envrc
```diff
@@ -2,7 +2,12 @@ if [ -f .env ]; then
     dotenv
 fi
 
-if ! command -v "devenv" >/dev/null; then
+expected_repo_name="$(awk -F'[][.]' '/^\[venv\./{print $3; exit}' devenv/config.ini 2>/dev/null)"
```
armenzg, line 5: When the worktree directory basename does not match the `devenv/config.ini` section (`[venv.sentry-protos]`), entering the directory triggers `devenv sync` via `.envrc` and crashes with a `KeyError`:

```text
$ cd /Users/.../wt-sentry-protos-supported-month-intervals
direnv: loading .../.envrc
Traceback (most recent call last):
  ...
  File ".../devenv/lib/venv.py", line 24, in get
    raise KeyError(f"section venv.{name} not found in repo config")
KeyError: 'section venv.wt-sentry-protos-supported-month-intervals not found in repo config'
```

dashed (approved):

PR #272, opened 2026-05-27
ci(python): Use .python-version as single source of truth

armenzg:
Use a tracked `.python-version` file as the single source of truth for Python selection across local dev and CI in `sentry-protos`, and fail fast when it drifts from `devenv/config.ini`.

**Single Source of Truth**

`.python-version` is now committed (and removed from `.gitignore`), and `build.yml`, `ci.yml`, and `docs.yml` all read Python via `python-version-file`. `actions/setup-python` is pinned to `v5.6.0`, which supports that input.

**Drift Protection**

A new `python-version-sync` CI job compares `.python-version` with `[venv.sentry-protos] python` in `devenv/config.ini` and fails the PR if they diverge.

**Packaging Contract**

`py/pyproject.toml` now sets `requires-python = ">=3.11"` so package metadata matches the pinned runtime.


.github/workflows/build.yml
```diff
@@ -14,9 +14,9 @@ jobs:
     steps:
       - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
 
-      - uses: actions/setup-python@e9aba2c848f5ebd159c070c61ea2c4e2b122355e # v2
+      - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
```
armenzg, line 17: This version can use `python-version-file`.

.github/workflows/build.yml
```diff
@@ -14,9 +14,9 @@ jobs:
...
         with:
-          python-version: 3.11
+          python-version-file: .python-version
```
armenzg, line 19: Shared Python version between local dev and CI.

.github/workflows/ci.yml
```diff
@@ -8,12 +8,30 @@ on:
       - 'buf.lock'
       - 'buf.gen.yaml'
       - 'build_rust/**'
+      - '.python-version'
+      - 'devenv/config.ini'
+      - '.github/workflows/**'
```
armenzg, line 13: The CI will trigger when these files are touched.

py/pyproject.toml
```diff
@@ -8,7 +8,7 @@ dynamic = ["version"]
 description = "Generated python code for sentry-protos"
 readme = "README.md"
 license = {file = "LICENSE.md"}
-requires-python = ">3.10"
+requires-python = ">=3.11"
```
armenzg, line 11: Making sure 3.10 would not be picked up by mistake.

Devenv only supports 3.11:
https://github.com/getsentry/sentry-protos/blob/6440f3b6fa0774d1b1b113b241dfb1d734eef957/devenv/config.ini#L7-L15

.python-version
```diff
@@ -0,0 +1 @@
+3.11.8
```
armenzg, line 1: Pinning the version.

dashed (approved):

PR #273, opened 2026-05-27
build(make): prefer .venv tooling for Python targets

armenzg:
Makefile Python targets now prefer tools from `.venv/bin` when a local virtualenv exists.

Previously, `build-py`, `test-py`, `docs`, `update-venv`, and `package-py` called bare `python`/`pip`/`pytest`, which could silently use an ambient interpreter. This updates those targets to resolve `.venv/bin/*` first and fall back to PATH only when `.venv` is not present.

That keeps local and CI-like workflows aligned with the project environment without breaking first-run setups.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,10 +3,10 @@
 
 # unstable protos are only included in local development and not part of release packages
 SENTRY_PROTOS_BUILD_UNSTABLE := 1
-PYTHON ?= $(if $(wildcard .venv/bin/python),.venv/bin/python,python)
-PIP ?= $(if $(wildcard .venv/bin/pip),.venv/bin/pip,pip)
-PYTEST ?= $(if $(wildcard .venv/bin/pytest),.venv/bin/pytest,pytest)
-SABLEDOCS ?= $(if $(wildcard .venv/bin/sabledocs),.venv/bin/sabledocs,sabledocs)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,23 +1,27 @@
...
 SENTRY_PROTOS_BUILD_UNSTABLE := 1
+PYTHON ?= $(if $(wildcard .venv/bin/python),.venv/bin/python,python)
+PIP ?= $(if $(wildcard .venv/bin/pip),.venv/bin/pip,pip)
+PYTEST ?= $(if $(wildcard .venv/bin/pytest),.venv/bin/pytest,pytest)
+SABLEDOCS ?= $(if $(wildcard .venv/bin/sabledocs),.venv/bin/sabledocs,sabledocs)
 
 .PHONY: update-venv
 update-venv:
-	pip install -r requirements.txt
+	$(PIP) install -r requirements.txt
```
markstory, line 13: Should we be switching over to `uv run` instead?

armenzg: Sounds good 👍🏻

armenzg: Good callout. I kept this PR on the existing `devenv`/`pip` flow to keep scope tight, but I pushed a follow-up commit that fixes the path issue by resolving `.venv` tools with absolute `$(CURDIR)` paths so subdirectory recipes work reliably.

If you want, I can open a separate PR to evaluate moving these Make targets to `uv run` end-to-end.

PR #276, opened 2026-05-27
feat: Add pending change service

noahsmartin:
Protos for the data service that holds pending changes. The kinds of pending changes can either be a change in package or a change in user config (payg budget or line items)


brendanhsentry (approved):

proto/sentry_protos/billing/v1/common/v1/pending_change.proto
```diff
@@ -0,0 +1,64 @@
+syntax = "proto3";
+
+package sentry_protos.billing.v1.common.v1;
+
+import "google/protobuf/empty.proto";
+
+// A pending change is a deferred subscription change for an organization that
+// will apply at the next billing period rollover. Each org has at most one
```
krithikravi, line 8: are we not going to support coterm changes then? for example, how would the use case of a user who hits their payg budget mid cycle, then wants to increase their budget to continue using sentry work?

proto/sentry_protos/billing/v1/common/v1/pending_change.proto
```diff
@@ -0,0 +1,64 @@
...
+// pending change at a time. Any field left unset means that aspect of the
+// subscription is not being changed.
+message PendingChange {
```
krithikravi, line 11: where do we plan to store these changes? if it's on the contract we can get away without an id, but if it is in a separate table, like in the existing system, we would need a unique identifier field as well

proto/sentry_protos/billing/v1/common/v1/pending_change.proto
noahsmartin: That wouldn't be through the "pending change" service, because it doesn't end up pending, but we will support it

proto/sentry_protos/billing/v1/common/v1/pending_change.proto
noahsmartin: A separate table, but we don't need an identifier in the proto right? There would be at most 1 per org so the protos can use the org id (which is in the request)

krithikravi (approved):

PR #279, opened 2026-05-28
feat(billing-platform): Support add-on line items

brendanhsentry:
part of https://linear.app/getsentry/issue/REVENG-122/support-add-ons

### DIFF
--- a/proto/sentry_protos/billing/v1/services/contract/v1/pricing_config.proto
+++ b/proto/sentry_protos/billing/v1/services/contract/v1/pricing_config.proto
@@ -68,6 +68,9 @@ message LineItemUids {
   repeated string uids = 1;
 }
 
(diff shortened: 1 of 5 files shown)


proto/sentry_protos/billing/v1/services/contract/v1/pricing_config.proto
```diff
@@ -68,13 +68,17 @@ message LineItemUids {
   repeated string uids = 1;
 }
 
+// For activating add-ons that do not necessarily require Reservation or PAYGBudget changes
+message Activation {}
```
brendanhsentry, line 72: chose an empty message because Seer doesn't have any special properties when "enabled". Containing it to a message will make it easier to make adjustments in the future

krithikravi (approved):

noahsmartin (approved):

PR #284, opened 2026-06-03
feat(billing): add tax id fields to billing details and contract requests

dashed:
## Summary

Adds the customer's tax registration number to the platform billing protos so contract invoicing can compute sales tax with the correct treatment (for example, EU reverse charge for tax-registered businesses).

## Context

This isn't a new concept — it exposes a value the billing system already stores per customer and already forwards to the tax provider:

- The customer's tax registration number (VAT / GST / …) is stored on `BillingDetails`, set and VAT-validated through the billing-details API.
- At tax-calculation time it is sent to the tax provider as the transaction's `businessIdentificationNo` (AvaTax), and a present tax id is what drives reverse-charge treatment for tax-registered businesses (e.g. in the EU/UK).

These fields add the two wire hops that were missing:

- `BillingDetails.tax_number` exposes the stored value over `GetBillingDetails`.
- `CreateContractRequest.customer_tax_id` / `RolloverContractRequest.customer_tax_id` let the caller forward it into contract creation and rollover, where tax is computed.

The two names mirror their respective domains: `tax_number` matches the persisted field, while `customer_tax_id` matches the tax-calculation input it feeds.

## Changes

| Message | File | New field |
| --- | --- | --- |
| `BillingDetails` | `billing_details/v1/billing_details.proto` | `optional string tax_number = 6` |
| `CreateContractRequest` | `contract/v1/endpoint_create_contract.proto` | `optional string customer_tax_id = 7` |
| `RolloverContractRequest` | `contract/v1/endpoint_rollover_contract.proto` | `optional string customer_tax_id = 6` |

`GetBillingDetailsResponse` already embeds `BillingDetails`, so it returns `tax_number` with no further change.

## Backwards compatibility

All three are new `optional` fields with previously-unused field numbers — additive only, with no field reuse, rename, or type change — so the change is backwards-compatible (`buf breaking` passes against `main`).

## Codegen

Rust bindings are regenerated and committed (`rust/src/…billing_details.v1.rs`, `…contract.v1.rs`). Python output is gitignored and regenerated by consumers.

REVENG-15

### DIFF
--- a/proto/sentry_protos/billing/v1/services/contract/v1/endpoint_rollover_contract.proto
+++ b/proto/sentry_protos/billing/v1/services/contract/v1/endpoint_rollover_contract.proto
@@ -19,7 +19,8 @@ message RolloverContractRequest {
   // pending change is being applied during this rollover.
   optional sentry_protos.billing.v1.common.v1.PendingChange pending_change = 5;
   // The customer's tax registration number (e.g. VAT ID), used to determine
-  // tax treatment for this contract. Unset if none is on file.
+  // tax treatment for this contract (for example, reverse charge for
+  // tax-registered businesses). Unset if none is on file.
   optional string customer_tax_id = 6;
 }
 
--- a/proto/sentry_protos/billing/v1/services/contract/v1/endpoint_rollover_contract.proto
+++ b/proto/sentry_protos/billing/v1/services/contract/v1/endpoint_rollover_contract.proto
@@ -18,6 +18,9 @@ message RolloverContractRequest {
   // The pending change to apply to the new contract, if any. Unset means no
   // pending change is being applied during this rollover.
   optional sentry_protos.billing.v1.common.v1.PendingChange pending_change = 5;
+  // The customer's tax registration number (e.g. VAT ID), used to determine
+  // tax treatment for this contract. Unset if none is on file.
+  optional string customer_tax_id = 6;
 }
 
 message RolloverContractResponse {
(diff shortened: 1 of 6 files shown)


proto/sentry_protos/billing/v1/services/contract/v1/endpoint_rollover_contract.proto
```diff
@@ -18,6 +18,10 @@ message RolloverContractRequest {
   // The pending change to apply to the new contract, if any. Unset means no
   // pending change is being applied during this rollover.
   optional sentry_protos.billing.v1.common.v1.PendingChange pending_change = 5;
+  // The customer's tax registration number (e.g. VAT ID), used to determine
+  // tax treatment for this contract (for example, reverse charge for
+  // tax-registered businesses). Unset if none is on file.
+  optional string customer_tax_id = 6;
```
brendanhsentry, line 24: Why do we need this for rollover contract? We pass the invoice line items through this request. So doesn't it make more sense to make use of the `customer_tax_id` when determining the line_items to pass to this request?

brendanhsentry (approved):
lgtm; small question

proto/sentry_protos/billing/v1/services/contract/v1/endpoint_rollover_contract.proto
dashed: @brendanhsentry im moving the tax calculation to upstream in the invoicer service rather than at the contract service. this changed the protos a bit. i'm preferring the protos field additions in https://github.com/getsentry/sentry-protos/pull/286 instead.

PR #290, opened 2026-06-05
feat(billing-platform): add InvoicerService.HandleChargeUpdated endpoint

armcknight:
Adds the proto definitions for a new `InvoicerService.HandleChargeUpdated`
endpoint, used by the Stripe `charge.updated` webhook to delegate to the
billing platform.

The shape mirrors the existing `HandleChargeSucceeded` and
`HandleChargeDisputed` endpoints: the request carries a `StripeCharge`
snapshot from the webhook payload, and the response signals via a
`handled` bool whether the billing platform owned and processed the
charge. When `handled=false`, the webhook caller falls through to the
legacy charge-sync path.

Prerequisite for the corresponding getsentry change in REVENG-113.

Refs REVENG-113


armcknight:
Closing — the consuming change (getsentry#20559) has dropped the `InvoicerService.HandleChargeUpdated` endpoint after review feedback. `charge.updated` is a pure sync event with no orchestration, so the webhook now calls `ChargeService.UpdateCharge` directly instead of going through a presentation-service wrapper. The `HandleChargeUpdated` proto endpoint is no longer needed.

If we ever need to re-introduce a wrapper (e.g. if `charge.updated` grows side effects on the platform side), we'll open a fresh proto PR at that point.

PR #294, opened 2026-06-05
ref(billing): added flexible price to shared line item pools

krithikravi:


proto/sentry_protos/billing/v1/services/package/v1/package.proto
brendanhsentry: we should probably deprecate this field too

brendanhsentry (approved):

PR #295, opened 2026-06-05
feat(billing): add invoice_item_types to LineItemDetails

dashed:
This is for https://github.com/getsentry/getsentry/pull/20573

## What

Adds a nested `InvoiceItemTypes` message to `line_item_details.proto` and attaches it to `LineItemDetails` as `invoice_item_types` (field number 2, previously unused).

```proto
message LineItemDetails {
  string uid = 1;
  InvoiceItemTypes invoice_item_types = 2;  // new
  string customer_facing_name = 3;
  string plural = 4;
  UnitInfo units = 5;
  BillableMetric billable_metric = 6;
}

message InvoiceItemTypes {
  optional string reserved  = 1;
  optional string ondemand  = 2;
  optional string activated = 3;
}
```

## Why

A single SKU's **reserved**, **on-demand**, and **activated** usage can each settle as a *distinct* `InvoiceItemType`. Collocating those identifiers on the SKU's `LineItemDetails` lets getsentry declare them per SKU, instead of growing a shared enum every time a new SKU/context combination appears.

Each context is `optional` because not every SKU uses every context — some SKUs (e.g. Seer) fold their usage irregularly and only populate a subset.

## Bindings

- Rust bindings regenerated (`make build-rust`) — the tracked `.rs` now contains `InvoiceItemTypes` and `LineItemDetails.invoice_item_types`.
- Python bindings regenerated (`make build-py`); the `py/` output is gitignored and not part of this diff.
- `Cargo.lock` left at main's version (the build's version-bump drift was reverted).

## Compatibility

Purely additive — a new message and a new field on a previously-unused field number. No existing fields change.

### DIFF
--- a/proto/sentry_protos/billing/v1/common/v1/line_item_details.proto
+++ b/proto/sentry_protos/billing/v1/common/v1/line_item_details.proto
@@ -10,6 +10,9 @@ message LineItemDetails {
   // Unique identifier for the line item.
   string uid = 1;
 
+  // InvoiceItemType identifiers for this SKU, keyed by billing context.
+  InvoiceItemTypes invoice_item_types = 2;
+
   // Customer-facing display name for the line item.
   string customer_facing_name = 3;
 
@@ -21,4 +24,20 @@ message LineItemDetails {
 
   // Defines how usage data is transformed into this line item.
   BillableMetric billable_metric = 6;
+}
+
+// InvoiceItemType identifiers for a single SKU, split by billing context.
+// Reserved, on-demand, and activated usage of the same SKU can each settle as a
+// distinct InvoiceItemType, so every context carries its own identifier instead
+// of collapsing into one. Contexts are optional because some SKUs (e.g. Seer)
+// fold their usage irregularly and do not populate every context.
+message InvoiceItemTypes {
+  // Type for reserved (prepaid) usage of this SKU.
+  optional string reserved = 1;
+
+  // Type for on-demand (pay-as-you-go) usage of this SKU.
+  optional string ondemand = 2;
+
+  // Type for activated usage of this SKU.
+  optional string activated = 3;
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -717,7 +717,7 @@ checksum = "955d28af4278de8121b7ebeb796b6a45735dc01436d898801014aced2773a3d6"
 
 [[package]]
 name = "sentry_protos"
-version = "0.23.0"
+version = "0.24.0"
 dependencies = [
  "prost",
  "prost-types",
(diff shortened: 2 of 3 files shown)


proto/sentry_protos/billing/v1/common/v1/line_item_details.proto
```diff
@@ -22,3 +25,19 @@ message LineItemDetails {
...
+// Reserved, on-demand, and activated usage of the same SKU can each settle as a
+// distinct InvoiceItemType, so every context carries its own identifier instead
+// of collapsing into one. Contexts are optional because some SKUs (e.g. Seer)
+// fold their usage irregularly and do not populate every context.
+message InvoiceItemTypes {
+  // Type for reserved (prepaid) usage of this SKU.
+  optional string reserved = 1;
+
+  // Type for on-demand (pay-as-you-go) usage of this SKU.
+  optional string ondemand = 2;
```
volokluev, line 39: according to the [changelog](https://app.notion.com/p/sentry/Billing-Platform-Changelog-3758b10e4b5d80508ac6d05c2ed56a8e?source=copy_link#3758b10e4b5d80589409f35136f7e8cc), ondemand has been replaced with payg, can we keep this consistent?

proto/sentry_protos/billing/v1/common/v1/line_item_details.proto
```diff
@@ -22,3 +25,19 @@ message LineItemDetails {
...
+
+  // Type for activated usage of this SKU.
+  optional string activated = 3;
```
volokluev, line 42: what is "activated usage"? what are the example values for it?

proto/sentry_protos/billing/v1/common/v1/line_item_details.proto
```diff
@@ -22,3 +25,19 @@ message LineItemDetails {
   // Defines how usage data is transformed into this line item.
   BillableMetric billable_metric = 6;
 }
+
+// InvoiceItemType identifiers for a single SKU, split by billing context.
```
volokluev, line 29: what is a SKU? Have we defined it somewhere?

proto/sentry_protos/billing/v1/common/v1/line_item_details.proto
```diff
@@ -22,3 +25,19 @@ message LineItemDetails {
...
+// Reserved, on-demand, and activated usage of the same SKU can each settle as a
+// distinct InvoiceItemType, so every context carries its own identifier instead
+// of collapsing into one. Contexts are optional because some SKUs (e.g. Seer)
+// fold their usage irregularly and do not populate every context.
+message InvoiceItemTypes {
```
volokluev, line 34: it's not clear to me why we would need to expose this in the proto as anything but an enum. We already have the `billable_metric`, the `customer_facing_name`, why would we also need to add `reserved_errors`, `reserved_trace_metric_bytes` etc to the list? Why not just an enum of 

```
enum InvoiceItemType {
    INVOICE_ITEM_TYPE_RESERVED = 1;
    INVOICE_ITEM_TYPE_PAYG = 2;
    INVOICE_ITEM_TYPE_ACTIVATED = 2;
}
```

I don't see why downstream services which need to generate a string like `reserved_errors` for their own purposes need that to be exposed in the lineitem proto

PR #300, opened 2026-06-09
ref(billing): Add flexible price to Reservation

skonves:
## Summary

- Deprecate `Reservation.reserved_price_cents` and add `flexible_reserved_price_cents: FlexiblePrice` (field 4) so reserved capacity on a contract can carry an annual-term override alongside the monthly base.
- Mirrors the existing FlexiblePrice migrations on `PackageConfig` (#288) and `SharedLineItemPool` (#294).
- This PR is schema-only; getsentry consumers are unchanged. Posting for the team to weigh in on the shape before wiring up annual reserved-line-item discounts.

## Why

REVENG-145 calls for applying the same annual discount we apply to the package base price to reserved line items. Today `Reservation.reserved_price_cents` is a flat `uint64` and is invoiced as-is in `create_invoice.subscription_line_items`, so there's no way to encode an annual `fixed` total alongside the monthly base.

After this lands, the resolver in `services/invoicer/pricing.py:compute_flexible_amount_cents` can be applied to reservations the same way it is for the package base.

## Test plan

- [ ] `buf lint` passes (ran via pre-commit on commit)
- [ ] `buf format` passes (ran via pre-commit on commit)
- [ ] CI rebuilds Python + Rust bindings cleanly
- [ ] Reviewers agree on naming (`flexible_reserved_price_cents`) and that putting the override on `Reservation` is the right home (alternative discussed: derive at invoice time from `LineItemConfig.reserved_rate` tiers on the package)

Refs REVENG-145

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -717,7 +717,7 @@ checksum = "955d28af4278de8121b7ebeb796b6a45735dc01436d898801014aced2773a3d6"
 
 [[package]]
 name = "sentry_protos"
-version = "0.26.0"
+version = "0.26.1"
 dependencies = [
  "prost",
  "prost-types",
--- a/proto/sentry_protos/billing/v1/services/contract/v1/pricing_config.proto
+++ b/proto/sentry_protos/billing/v1/services/contract/v1/pricing_config.proto
@@ -4,6 +4,7 @@ package sentry_protos.billing.v1.services.contract.v1;
 
 import "google/protobuf/empty.proto";
 import "google/protobuf/timestamp.proto";
+import "sentry_protos/billing/v1/common/v1/flexible_price.proto";
 import "sentry_protos/billing/v1/services/contract/v1/billing_config.proto";
 import "sentry_protos/billing/v1/services/contract/v1/sku.proto";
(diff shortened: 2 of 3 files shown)


krithikravi (commented):
do we need to make this change to the reservation proto? i think a contract can either be annual or monthly at any given point, so can't we just check the contract's interval, number of reserved units, and the package monthly/annual reserved costs during checkout, and  calculate+store the appropriate monthly/annual cost in the existing reserved_price_cents field? we can also carry out this operation when invoicing, since we are calling the package service anyway for the package price. it seems like this new flexible price field will just store the exact same information as the package's reserved_rate field, not sure if we need to duplicate that.

skonves:
Closing per reviewer feedback (thanks @ everyone on the call). The new ``flexible_reserved_price_cents`` field would duplicate information already encoded in ``LineItemConfig.reserved_rate.tiers[*].price``. Instead, the getsentry invoicer will look up the matching tier from the package at invoice time and use ``compute_flexible_amount_cents`` on its ``FlexiblePrice``.

Follow-up landing in getsentry/getsentry#20626 — no proto change required.

Refs REVENG-145

PR #309, opened 2026-06-12
chore(billing-platform): Active trials endpoint

brendanhsentry:
protos part of https://linear.app/getsentry/issue/REVENG-206/get-active-trials-service-method


proto/sentry_protos/billing/v1/services/engagement/v1/endpoint_get_active_trials.proto
```diff
@@ -0,0 +1,33 @@
+syntax = "proto3";
+
+package sentry_protos.billing.v1.services.engagement.v1;
+
+import "sentry_protos/billing/v1/date.proto";
+
+message TrialConfig {
```
krithikravi, line 7: how would we handle feature entitlements for trials? would that be stored in these models or elsewhere?

brendanhsentry: They can be. But feature entitlements are not in scope for the Engagement service

krithikravi (approved):
lgtm

PR #303, opened 2026-06-11
feat(billing-platform): add refund types + record_charge_refunds endpoint (REVENG-157)

armcknight:
## Summary

Shared refund types + the endpoint driven by the `charge.refunded` webhook flow:

- **`common/v1/stripe_charge.proto`** — new `StripeRefund` message + `repeated refunds` field on `StripeCharge`. Webhook handlers need per-refund metadata (id, amount, reason) for idempotent ingestion; the aggregate `amount_refunded` on the charge alone isn't enough.
- **`services/charge/v1/charge.proto`** — new `PlatformRefund` canonical projection (parallels `PlatformCharge`), embedded as `repeated refunds` on `PlatformCharge`. The previously-cached aggregate scalars `refunded` (field 7) and `amount_refunded` (field 8) stay in the message marked `[deprecated = true]` (per `sentry-protos` policy that disallows field deletion) and consumers should derive both from the `refunds` list — `len(refunds) > 0` for "any refund happened" and `sum(refunds[*].amount_cents)` for the total.
- **`services/charge/v1/endpoint_record_charge_refunds.proto`** (new) — `ChargeService.record_charge_refunds`, called directly by the `charge.refunded` webhook layer to record per-refund rows idempotently and return the charge with its refund list embedded.

## Proto stack

| Proto PR | Paired getsentry PR | Contents |
|---|---|---|
| **#303 (this PR, merged)** | [getsentry#20611](https://github.com/getsentry/getsentry/pull/20611) | `StripeRefund` + `PlatformRefund` shared types, `record_charge_refunds` |
| #305 | [getsentry#20613](https://github.com/getsentry/getsentry/pull/20613) | `endpoint_list_refunds.proto` (for customer + admin refund rendering) |

## Test plan

- [x] Rust bindings regenerated by CI's `Build Rust bindings` step + auto-committed
- [x] Python bindings regenerated by CI
- [x] Consumed downstream by getsentry#20611

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #312, opened 2026-06-15
feat(billing-platform): add batched ChargeService.list_charges_for_invoice_ids (REVENG-157)

armcknight:
## Summary

Adds a batched endpoint that returns `PlatformCharge` (the canonical projection, with refunds embedded via `PlatformCharge.refunds` from [#303](https://github.com/getsentry/sentry-protos/pull/303)) for every charge tied to any of a list of invoices.

- **`services/charge/v1/endpoint_list_charges.proto`** — new `ListChargesForInvoiceIdsRequest` (`repeated uint64 invoice_ids`) + `ListChargesForInvoiceIdsResponse` (flat list of `PlatformCharge` ordered by `(invoice_id, date_added)`).

## Why

[getsentry#20613](https://github.com/getsentry/getsentry/pull/20613) hit a Cursor-bot-flagged N+1 in the customer invoices list — paginated rendering was calling `ChargeService.list_refunds_by_invoice` once per invoice. Earlier review feedback from Noah ([getsentry#20611 thread](https://github.com/getsentry/getsentry/pull/20611#discussion_r3397952041)) pointed at the same pattern: derive aggregates from per-row data, batch where possible.

This PR went with **charges-with-embedded-refunds** rather than a separate `list_refunds_by_invoice_ids` because:

1. **It leverages existing proto surface.** `PlatformCharge.refunds` already exists (added in #303); we just need a batch method that populates and returns it.
2. **It sets up admin surfaces.** Admin invoice details (deferred follow-up from getsentry#20613) wants full charge + refund detail in one round trip — this is the right shape.
3. **It still solves the customer-list case.** Consumers needing only refund totals sum `charge.refunds[*].amount_cents` grouped by `charge.invoice_id`.

The singleton `ListChargesForInvoiceRequest` / slim `Charge` type in the same file (dormant, defined in [#211](https://github.com/getsentry/sentry-protos/pull/211) but never implemented) is left as-is — deferring its deprecation to a separate proto-policy follow-up.

## Test plan

- [ ] Rust + Python bindings regenerated by CI
- [ ] `buf breaking` passes (verified locally — additive only)
- [ ] Once released + sentry pin bumped, consumed by getsentry#20613's `_PlatformInvoiceSource` to eliminate the N+1

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -717,7 +717,7 @@ checksum = "955d28af4278de8121b7ebeb796b6a45735dc01436d898801014aced2773a3d6"
 
 [[package]]
 name = "sentry_protos"
-version = "0.30.0"
+version = "0.30.1"
 dependencies = [
  "prost",
  "prost-types",
--- a/proto/sentry_protos/billing/v1/services/charge/v1/endpoint_list_charges.proto
+++ b/proto/sentry_protos/billing/v1/services/charge/v1/endpoint_list_charges.proto
@@ -2,6 +2,8 @@ syntax = "proto3";
 
 package sentry_protos.billing.v1.services.charge.v1;
 
+import "sentry_protos/billing/v1/services/charge/v1/charge.proto";
+
 message Charge {
   uint64 amount_cents = 1;
   bool paid = 2;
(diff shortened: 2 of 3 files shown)


armcknight:
Closing — switching the batch shape to a dataclass prototype on the getsentry side (per the billing-platform CLAUDE.md prototyping convention). We'll solidify the wire contract here after the design has been validated in production. The current PlatformCharge.refunds field (from #303) is sufficient for the dataclass implementation.

PR #313, opened 2026-06-15
feat(billing): Added retry charge endpoint protos

krithikravi:
This endpoint will allow us to mark an invoice as `needs_charged=True` and `next_payment_attempt=now` so we can port the admin payment retry endpoint to the billing platform.

### DIFF
--- a/proto/sentry_protos/billing/v1/services/contract/v1/endpoint_retry_charge.proto
+++ b/proto/sentry_protos/billing/v1/services/contract/v1/endpoint_retry_charge.proto
@@ -3,7 +3,7 @@ syntax = "proto3";
 package sentry_protos.billing.v1.services.contract.v1;
 
 message RetryChargeRequest {
-  uint64 invoice_id = 1;
+  string invoice_guid = 1;
 }
 
 message RetryChargeResponse {
--- a/proto/sentry_protos/billing/v1/services/contract/v1/endpoint_retry_charge.proto
+++ b/proto/sentry_protos/billing/v1/services/contract/v1/endpoint_retry_charge.proto
@@ -1,0 +1,13 @@
+syntax = "proto3";
+
+package sentry_protos.billing.v1.services.contract.v1;
+
+import "sentry_protos/billing/v1/services/contract/v1/invoice.proto";
+
+message RetryChargeRequest {
+  Invoice invoice = 1;
+}
+
+message RetryChargeResponse {
+  Invoice invoice = 1;
+}
--- a/proto/sentry_protos/billing/v1/services/contract/v1/endpoint_retry_charge.proto
+++ b/proto/sentry_protos/billing/v1/services/contract/v1/endpoint_retry_charge.proto
@@ -2,12 +2,10 @@ syntax = "proto3";
 
 package sentry_protos.billing.v1.services.contract.v1;
 
-import "sentry_protos/billing/v1/services/contract/v1/invoice.proto";
-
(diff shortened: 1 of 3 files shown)


proto/sentry_protos/billing/v1/services/contract/v1/endpoint_retry_charge.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto3";
+
+package sentry_protos.billing.v1.services.contract.v1;
+
+import "sentry_protos/billing/v1/services/contract/v1/invoice.proto";
+
+message RetryChargeRequest {
+  Invoice invoice = 1;
```
volokluev: why do we need to pass in the entire invoice object?

brendanhsentry: do we need the whole invoice in the request? would the invoice_id be enough?

proto/sentry_protos/billing/v1/services/contract/v1/endpoint_retry_charge.proto
```diff
@@ -0,0 +1,13 @@
...
+}
+
+message RetryChargeResponse {
+  Invoice invoice = 1;
```
brendanhsentry: again, do we need the whole invoice in the response? would `needs_charge` and `next_payment_attempt` be sufficient

krithikravi: i'll switch to a boolean, the reason we had the invoice object here was because the legacy endpoint called a method that directly returned the invoice object, but we can just use get_invoice here anyway

proto/sentry_protos/billing/v1/services/contract/v1/endpoint_retry_charge.proto
krithikravi: not necessary, refactored

proto/sentry_protos/billing/v1/services/contract/v1/endpoint_retry_charge.proto
volokluev: I actually think returning the entire invoice makes sense. If you're trying to save on bits all fields in a protobuf are technically optional but I think it makes sense that you retry the charge of an invoice id, you get back the resulting object

brendanhsentry (approved):
looking closer at the endpoint, you might want to have invoice_guid instead of invoice_id in the request but i'll let you decide

proto/sentry_protos/billing/v1/services/contract/v1/endpoint_retry_charge.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto3";
+
+package sentry_protos.billing.v1.services.contract.v1;
+
+message RetryChargeRequest {
+  uint64 invoice_id = 1;
+}
+
+message RetryChargeResponse {
+  bool updated = 1;
```
volokluev, line 10: I commented on an outdated comment but I'll comment again:

I actually think returning the entire invoice makes sense. You retry the charge of an invoice id, you get back the resulting object, it's a very common database pattern for CRUD operations

krithikravi: tbh we can go either way, i'll stick with just the boolean to match the existing pattern of `MarkInvoicePaidResponse`

PR #315, opened 2026-06-17
chore(billing-platform): Add protos for set_pending_tax_transaction endpoint

dashed:
Adds the `SetPendingTaxTransaction` request/response messages to the contract service for the PR https://github.com/getsentry/getsentry/pull/20624.

`SetPendingTaxTransactionRequest` carries the `invoice_id` and the tax provider's `external_reference`; `SetPendingTaxTransactionResponse` returns whether a matching invoice was updated. These back the getsentry contract service's `set_pending_tax_transaction`, which records the provider-side tax document reference on an invoice as a pending tax transaction so it can be committed or voided once the charge outcome is known. That method is currently prototyped with proto-free dataclasses (the dataclass-prototype pattern); these messages are what it migrates to.

Message-only and consistent with the other contract endpoints — `uint64 invoice_id`, bare `string`, one `endpoint_*.proto` for the pair, mirroring `MarkInvoicePaid`. Generated stubs are produced at build time, so only the `.proto` source is included here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -717,7 +717,7 @@ checksum = "955d28af4278de8121b7ebeb796b6a45735dc01436d898801014aced2773a3d6"
 
 [[package]]
 name = "sentry_protos"
-version = "0.30.1"
+version = "0.31.2"
 dependencies = [
  "prost",
  "prost-types",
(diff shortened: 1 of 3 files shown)


dashed:
Superseded by #319. The atomic-linkage refactor (getsentry getsentry/getsentry#20754) folds the tax-document linkage into the contract create/rollover requests (`tax_transaction_code` on CreateContractRequest + RolloverContractRequest) and removes the standalone `set_pending_tax_transaction` endpoint, so these standalone messages have no consumer.

PR #318, opened 2026-06-18
ref: Add supports on demand

noahsmartin:
Not all contracts support on demand. Credit card + an override for enterprise do, we can abstract that implementation to just the contract service and check a boolean flag like this everywhere else

### DIFF
--- a/proto/sentry_protos/billing/v1/services/contract/v1/billing_config.proto
+++ b/proto/sentry_protos/billing/v1/services/contract/v1/billing_config.proto
@@ -61,4 +61,9 @@ message BillingConfig {
   // The number-of-months interval the contract was signed under
   // (1 = monthly, 12 = annual). Frozen for the life of the contract.
   uint32 month_interval = 7;
+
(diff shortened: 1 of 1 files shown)


proto/sentry_protos/billing/v1/services/contract/v1/billing_config.proto
```diff
@@ -61,4 +61,9 @@ message BillingConfig {
   // The number-of-months interval the contract was signed under
   // (1 = monthly, 12 = annual). Frozen for the life of the contract.
   uint32 month_interval = 7;
+
+  // Whether the org is allowed to incur on-demand (pay-as-you-go) usage.
+  // Credit-card orgs always support on-demand; invoiced orgs that should
+  // support it are an explicit override.
+  bool supports_on_demand = 8;
```
brendanhsentry: should we call it `supports_payg` to avoid keeping the legacy "on_demand" terminology around

brendanhsentry (approved):

PR #322, opened 2026-06-22
ref: Add protos fields for customer subscription API

noahsmartin:
These fields will be needed as part of the customer details API

### DIFF
--- a/proto/sentry_protos/billing/v1/services/contract/v1/billing_config.proto
+++ b/proto/sentry_protos/billing/v1/services/contract/v1/billing_config.proto
@@ -66,4 +66,12 @@ message BillingConfig {
   // Credit-card orgs always support payg; invoiced orgs that should
   // support it are an explicit override.
   bool supports_payg = 8;
+
+  // Whether this is a managed (sales-assisted) subscription rather than
+  // self-serve.
+  bool is_managed = 9;
+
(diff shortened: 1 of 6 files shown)


proto/sentry_protos/billing/v1/services/contract/v1/billing_config.proto
```diff
@@ -66,4 +66,12 @@ message BillingConfig {
...
   // support it are an explicit override.
   bool supports_payg = 8;
+
+  // Whether this is a managed (sales-assisted) subscription rather than
+  // self-serve.
+  bool is_managed = 9;
+
+  // Whether the org has a soft cap: usage past reserved volume is allowed
+  // (and billed) instead of hard-stopping ingestion.
+  bool has_soft_cap = 10;
```
armenzg, line 76: Do you know if we intend to preserve soft caps? or if we currently use it?

armenzg (approved):

proto/sentry_protos/billing/v1/services/contract/v1/billing_config.proto
noahsmartin: It is used so I think we should try and support it, would operate on the assumption that we are supporting it for now and can re-evaluate later if it needs to be dropped

PR #326, opened 2026-06-23
Add allow_payg field to PackageConfig

noahsmartin:
Add a boolean field indicating whether a package permits pay-as-you-go (PAYG) billing, so consumers can gate PAYG behavior per package.


brendanhsentry (commented):
I think PAYG is only unavailable for developer plans. So I'm not sure this field is necessary

brendanhsentry (approved):

PR #331, opened 2026-06-25
ref(billing-platform): Split up monetary and unit grants

brendanhsentry:
Unit and monetary grants have vastly different return shapes and use cases. It makes more sense to separate them from a single ActiveGrant message to avoid porting unnecessary fields to both grant types.

Also deprecates a bunch of messages that will not be used.

### DIFF
--- a/proto/sentry_protos/billing/v1/services/engagement/v1/endpoint_get_monetary_grants.proto
+++ b/proto/sentry_protos/billing/v1/services/engagement/v1/endpoint_get_monetary_grants.proto
@@ -1,0 +1,30 @@
+syntax = "proto3";
+
+package sentry_protos.billing.v1.services.engagement.v1;
+
+enum EffectiveMonetaryType {
+  EFFECTIVE_MONETARY_TYPE_UNSPECIFIED = 0;
+  EFFECTIVE_MONETARY_TYPE_CENTS = 1;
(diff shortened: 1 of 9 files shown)


proto/sentry_protos/billing/v1/services/engagement/v1/endpoint_get_monetary_grants.proto
```diff
@@ -0,0 +1,30 @@
...
+  EFFECTIVE_MONETARY_SOURCE_UNSPECIFIED = 0;
+  EFFECTIVE_MONETARY_SOURCE_BALANCE = 1;
+  EFFECTIVE_MONETARY_SOURCE_RECURRING_CREDIT = 2;
+}
+
+message EffectiveMonetaryGrant {
+  // PERCENT amounts are percentage points: 10 = 10%
+  uint64 amount = 1;
+  EffectiveMonetaryType type = 2;
+  EffectiveMonetarySource source = 3;
```
brendanhsentry, line 21: added `source` to `EffectiveMonetaryGrant` because we'll need to distinguish between discounts that resulted from recurring credits vs how much balance was used.

Also dropped start and end dates bc monetary grants are applied to contracts at invoicing time.

krithikravi (approved):

PR #339, opened 2026-06-30
chore(billing-platform): Convert BalanceService dataclasses in protobufs

brendanhsentry:
Turns [dataclasses](https://github.com/getsentry/getsentry/blob/cc4d614c083f0875c5a2d3ef4f7923a1a232ac65/getsentry/billing/platform/services/engagement/services/balance/service.py#L17-L70) into protobufs


proto/sentry_protos/billing/v1/services/engagement/v1/endpoint_assign_balance_contract.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto3";
+
+package sentry_protos.billing.v1.services.engagement.v1;
+
+message AssignBalanceContractRequest {
```
volokluev, line 5: what does this actually do?

volokluev (commented):
These names are not self explanatory, please add some documentation as to what these endpoints actually do

proto/sentry_protos/billing/v1/services/engagement/v1/endpoint_assign_balance_contract.proto
brendanhsentry: yeah you're right this is pretty bad. i have a [PR](https://github.com/getsentry/getsentry/pull/20888) to refactor in getsentry.

after that's merged, i'll come back and fix these protobuf messages

PR #350, opened 2026-07-08
feat(billing): Add PackageOption/PackageCatalog domain model

skonves:
Billing Platform domain-model protos for the package catalog, plus a package-scoped rate card endpoint.

- `PackageOption` / `PackageCatalog` and the `GetPackageCatalog` / `GetPackages` endpoints — the catalog of packages a customer can move to, and how to resolve options to their `PackageConfig`s. `AcquisitionMode` is `UNSPECIFIED` / `SELF_SERVE`.
- `GetRateCardForPackage` — a new endpoint (`package_uid` + `month_interval`) that builds a package's list-price rate card **without a contract**. `GetRateCardRequest` is left untouched, so this is purely additive — no breaking change.

### DIFF
--- a/proto/sentry_protos/billing/v1/services/package/v1/package_catalog.proto
+++ b/proto/sentry_protos/billing/v1/services/package/v1/package_catalog.proto
@@ -2,13 +2,10 @@ syntax = "proto3";
 
 package sentry_protos.billing.v1.services.package.v1;
 
-// How a customer can start on a package: sign up on their own, work with
-// sales, or receive it as a grant (e.g. open source / nonprofit).
+// How a customer can start on a package.
 enum AcquisitionMode {
   ACQUISITION_MODE_UNSPECIFIED = 0;
   ACQUISITION_MODE_SELF_SERVE = 1; // customer can sign up on their own
-  ACQUISITION_MODE_SALES = 2; // requires working with sales
-  ACQUISITION_MODE_SPONSORED = 3; // granted (open source / nonprofit)
 }
 
 // A package a customer can subscribe to, such as the Team or Business package.
--- a/proto/sentry_protos/billing/v1/services/package/v1/package_catalog.proto
+++ b/proto/sentry_protos/billing/v1/services/package/v1/package_catalog.proto
@@ -1,0 +1,47 @@
+syntax = "proto3";
+
+package sentry_protos.billing.v1.services.package.v1;
+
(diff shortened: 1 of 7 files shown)


proto/sentry_protos/billing/v1/services/package/v1/package_catalog.proto
```diff
@@ -0,0 +1,47 @@
+syntax = "proto3";
+
+package sentry_protos.billing.v1.services.package.v1;
+
+// How a customer can start on a package: sign up on their own, work with
+// sales, or receive it as a grant (e.g. open source / nonprofit).
+enum AcquisitionMode {
```
noahsmartin, line 6: I would keep this as only self_serve for now. It's not clear to me why we would need separate ones for both sales and sponsored rather than just one for sales (sales also makes packages sponsored). So not speculating now and just include what we know we need seems like it would be better

proto/sentry_protos/billing/v1/services/package/v1/package_catalog.proto
```diff
@@ -0,0 +1,47 @@
...
+}
+
+// A package a customer can subscribe to, such as the Team or Business package.
+message PackageOption {
+  // Unique identifier for this specific package. Use it with GetPackage to
+  // fetch the package's full details.
+  string uid = 1;
+
+  // The package's display name, e.g. "Team". Shown to customers as-is.
+  string display_name = 2;
```
noahsmartin: Why not use the existing package display name field? What would it mean if this was different from the package display name?

proto/sentry_protos/billing/v1/services/package/v1/package_catalog.proto
```diff
@@ -0,0 +1,47 @@
...
+  // talking to sales, flagged via AcquisitionMode).
+  repeated PackageOption packages = 1;
+
+  // uid of the free option within `packages`. Being on it means having no paid
+  // subscription — a distinct state — so it is called out for consumers that
+  // treat it specially.
+  string free_package_uid = 2;
+
+  // uid of the default option within `packages`: the one checkout pre-selects.
+  string default_package_uid = 3;
```
noahsmartin, line 38: I like the idea but I doubt the frontend supports this for now. To avoid extra churn in the protos I'd recommend removing it for now and we can add it later if we need it

noahsmartin (approved):

PR #352, opened 2026-07-08
feat(billing): add charge to CaptureChargeResponse

armcknight:
## Summary

- Adds an optional `PlatformCharge` field on `CaptureChargeResponse` (tag 3) so callers of `capture_charge` can consume the persisted charge row directly.
- Mirrors the shape of `UpdateChargeResponse.charge` (same optional `PlatformCharge`), so downstream code paths that need the full Charge projection have one uniform surface.
- Backward compatible: existing producers that don't set the field leave it unset; existing consumers that don't read the field are unaffected.

## Motivation

In getsentry PR [#20833](https://github.com/getsentry/getsentry/pull/20833), the platform Pay-Now webhook path does:

```py
capture_response = ChargeService().capture_charge(CaptureChargeRequest(...))
update_response = ChargeService().update_charge(UpdateChargeRequest(stripe_charge=...))
charge = update_response.charge
```

The second call exists only because `CaptureChargeResponse` doesn't currently carry the persisted charge -- so we have to round-trip through `update_charge` to get the `PlatformCharge` proto the downstream code (e.g. `charge.HasField("invoice_id")`, `charge.paid`) needs.

Reviewer feedback on that PR asked why we call `update` immediately after `capture` given we already provided the `stripe_charge` to `capture`; with this proto change the getsentry side can consume `capture_response.charge` and drop the redundant `update_charge` round-trip.

## Test plan

- [ ] `make build-py` produces `endpoint_capture_charge_pb2.pyi` with the new field
- [ ] Import + descriptor check: `CaptureChargeResponse.DESCRIPTOR.fields` includes `charge` at tag 3

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -717,7 +717,7 @@ checksum = "955d28af4278de8121b7ebeb796b6a45735dc01436d898801014aced2773a3d6"
 
 [[package]]
 name = "sentry_protos"
-version = "0.41.1"
+version = "0.42.0"
 dependencies = [
  "prost",
  "prost-types",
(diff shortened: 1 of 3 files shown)


armcknight:
Closing -- addressing this via caller-side restructure in getsentry#20833 instead. The redundant update_charge after capture_charge is being replaced by a direct read of request.stripe_charge / int(request.invoice_id), which are guaranteed equivalent to charge.paid / charge.invoice_id in the Pay Now flow. No proto change needed.

PR #354, opened 2026-07-10
chore(billing-platform): Add is_sponsored field to PackageConfig

brendanhsentry:
This will be used to display certain things in the UI and the EngagementService will use it to gate trial eligibility

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -717,7 +717,7 @@ checksum = "955d28af4278de8121b7ebeb796b6a45735dc01436d898801014aced2773a3d6"
 
 [[package]]
 name = "sentry_protos"
-version = "0.42.2"
+version = "0.43.0"
 dependencies = [
  "prost",
  "prost-types",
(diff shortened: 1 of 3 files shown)


volokluev:
didn't @skonves add the acquisition_channel enum recently? Isn't this where it would go?

brendanhsentry:
> didn't @skonves add the acquisition_channel enum recently? Isn't this where it would go?

@volokluev yeah that is a good point. So instead of a boolean field we can add an `acquisition_mode` to the package config (and deprecate the `is_enterprise` field too).

@skonves do you have any thoughts on this?

skonves:
I did, but last minute we decided we'd pull out the concept of sponsored because there is additional necessary complexity. Part of the concern was that _admin could theoretically create a contract with any package and have it sponsored for a given reason. For this reason, we removed it from the catalog response.

volokluev:
Do does his mean a package can both be self-serve and sponsored? That's what this PR would imply and it doesn't seem right to me. The enum still seems like the right place for it.

brendanhsentry:
synced offline and are going in a different direction

PR #355, opened 2026-07-10
chore(billing-platform): Add SponsorshipConfig to Contract

brendanhsentry:
take 2 of https://github.com/getsentry/sentry-protos/pull/354


skonves (approved):
I actually really like this. It answers the question of "who's paying for this?" which is naturally a billing concept.

PR #366, opened 2026-07-16
ref(billing-platform): Add source and source_id when creating RecurringCredits

brendanhsentry:
follow up to https://github.com/getsentry/getsentry/pull/21102

### DIFF
--- a/proto/sentry_protos/billing/v1/services/engagement/v1/endpoint_create_recurring_credit.proto
+++ b/proto/sentry_protos/billing/v1/services/engagement/v1/endpoint_create_recurring_credit.proto
@@ -13,9 +13,7 @@ message CreateRecurringCreditRequest {
   uint64 number_of_periods = 5;
   uint64 contract_id = 6;
 
-  optional CreditSource source = 7;
-  // Identifier of the record within source (ie a PromoCode id).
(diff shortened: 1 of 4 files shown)


proto/sentry_protos/billing/v1/services/engagement/v1/endpoint_create_recurring_credit.proto
```diff
@@ -12,6 +12,10 @@ message CreateRecurringCreditRequest {
   uint64 amount = 4;
   uint64 number_of_periods = 5;
   uint64 contract_id = 6;
+
+  optional CreditSource source = 7;
+  // Identifier of the record within source (ie a PromoCode id).
+  optional string source_id = 8;
```
volokluev: nit: I'd put these in a single message rather than two separate fields to signify that these should come together as a package

volokluev (approved):

proto/sentry_protos/billing/v1/services/engagement/v1/endpoint_create_recurring_credit.proto
brendanhsentry: good idea

PR #368, opened 2026-07-17
feat(billing): Add common retention policy types

dashed:
Closes https://linear.app/getsentry/issue/REVENG-333/introduce-retention-protos

This introduces common and re-usable retention protos that'll be used in:
- contract (customer specific retention overrides)
- package (default retentions for a given pricing package)

Prior art from legacy billing system that's ported into new billing platform world:
- `TieredPlanItem.retention` and `TieredPlanItem.downsampled_event_retention`
https://github.com/getsentry/getsentry/blob/b9f4cddb1d463c5f56083cad5b1201f165f52ccd/getsentry/billing/plans/tiered_plan_item.py#L38-L39
- `BillingMetricHistory.retention_days` and `BillingMetricHistory.retention_days_downsampled` https://github.com/getsentry/getsentry/blob/b9f4cddb1d463c5f56083cad5b1201f165f52ccd/getsentry/models/billingmetrichistory.py#L491-L493

### DIFF
--- a/proto/sentry_protos/billing/v1/common/v1/retention.proto
+++ b/proto/sentry_protos/billing/v1/common/v1/retention.proto
@@ -4,28 +4,34 @@ package sentry_protos.billing.v1.common.v1;
 
 import "sentry_protos/billing/v1/data_category.proto";
 
-// The complete retention settings for a data category.
+// Retention settings for one data category.
 //
-// These are complete package/common settings, not sparse override semantics.
+// A package uses this message for a complete policy. Do not use this message
+// for sparse updates or overrides. Services validate categories, apply
+// precedence, preserve legacy behavior, and calculate effective retention.
 message RetentionSettings {
-  // Standard/full-fidelity retention in calendar days. Must be greater than zero.
+  // The number of calendar days to retain standard data. The value must be
+  // greater than zero.
   uint32 standard_days = 1;
 
-  // Retention for a distinct downsampled representation in calendar days.
+  // The number of calendar days to retain downsampled data.
   //
-  // Absence means the category has no distinct downsampled representation. A
-  // present zero is a package compatibility value interpreted by the retention
-  // resolver as the effective standard retention. A positive value is a
-  // concrete downsampled retention duration.
+  // If this field is absent, the category has no separate downsampled data. If
+  // the value is zero, the resolver uses the effective standard retention. This
+  // value preserves legacy behavior. A positive value is a concrete duration.
   optional uint32 downsampled_days = 2;
 }
 
-// Associates complete retention settings with a billing data category.
+// Complete retention settings for one billing data category.
+//
+// Retention is keyed by billing data category, not by line item or SKU. A line
+// item can bill more than one category. A category can be billed by more than
+// one line item. Some line items have no retention. A shared pool bills no
+// single category. Do not read a category from a line item.
 message DataCategoryRetention {
-  // The supported billing data category. DATA_CATEGORY_UNSPECIFIED and
-  // DATA_CATEGORY_UNKNOWN are invalid.
+  // Do not use DATA_CATEGORY_UNSPECIFIED or DATA_CATEGORY_UNKNOWN.
   sentry_protos.billing.v1.DataCategory category = 1;
 
-  // The category's retention settings. Required by the domain model.
+  // The category retention settings. This message is required.
   RetentionSettings settings = 2;
 }
--- a/proto/sentry_protos/billing/v1/common/v1/retention.proto
+++ b/proto/sentry_protos/billing/v1/common/v1/retention.proto
@@ -1,0 +1,31 @@
+syntax = "proto3";
+
+package sentry_protos.billing.v1.common.v1;
+
+import "sentry_protos/billing/v1/data_category.proto";
+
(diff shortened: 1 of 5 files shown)


proto/sentry_protos/billing/v1/common/v1/retention.proto
```diff
@@ -0,0 +1,51 @@
...
+
+import "sentry_protos/billing/v1/data_category.proto";
+
+// The standard retention period for stored data.
+//
+// This message is required by the domain model, although proto3 cannot enforce
+// message presence.
+message StandardRetention {
+  // The number of calendar days that data is retained. Must be greater than zero.
+  uint32 days = 1;
```
dashed: This maps to `TieredPlanItem.retention`. 

See: https://github.com/getsentry/getsentry/blob/b9f4cddb1d463c5f56083cad5b1201f165f52ccd/getsentry/billing/plans/tiered_plan_item.py#L38

proto/sentry_protos/billing/v1/common/v1/retention.proto
```diff
@@ -0,0 +1,51 @@
...
+}
+
+// Indicates that downsampled data uses the standard retention period.
+message InheritStandard {}
+
+// The retention policy for a category's downsampled data.
+message DownsampledRetention {
```
dashed: Maps to `TieredPlanItem.downsampled_event_retention`

see: https://github.com/getsentry/getsentry/blob/b9f4cddb1d463c5f56083cad5b1201f165f52ccd/getsentry/billing/plans/tiered_plan_item.py#L39

proto/sentry_protos/billing/v1/common/v1/retention.proto
```diff
@@ -0,0 +1,52 @@
...
+//
+// This message describes the whole policy, not a sparse update or override.
+message RetentionSettings {
+  // Retention for the standard, full-fidelity representation. An absent message
+  // is invalid.
+  StandardRetention standard = 1;
+
+  // Retention for a distinct downsampled representation. Absence means that no
+  // such representation exists; it does not mean to inherit standard retention.
+  optional DownsampledRetention downsampled = 2;
```
dashed: Note that, it is legal for a data category to have both standard and downsampled retention settings.

For example, spans have retention of 30 days, while its downsampled counter part can be stored up to 13 months on AM3 business plans: https://github.com/getsentry/getsentry/blob/ac0ba4af432a399b77b42227d4017eaa14be65e4/getsentry/billing/plans/am3/business.py#L173-L178

noahsmartin: I don't think I understand this comment. The standard retention setting is not optional in the proto but the downsampled one is optional. Doesn't that mean some categories would have both (the optional one would be non-null)

dashed: @noahsmartin yep. for example, logs, transactions (am2), and spans would have both.

https://docs.sentry.io/security-legal-pii/security/data-retention-periods/#sampled-retention-for-business-and-enterprise-plans

there's also this notion doc i found that talks abit about this https://app.notion.com/p/sentry/Data-Retention-29a8b10e4b5d8037b796ecfb8858ab17

volokluev (commented):
I think this is more confusing than it needs to be. The interface is leaking business logic. 

It seems like retention settings can specify:

1. standard retention (retention for unsampled data)
2. downsampled retetion (for donwsampled data)

from my perspective, a message which exposes those two numbers should be the only thing exposed in the protos. The services which will use these message structures will have the policy decisions specified in their code, there is no need to specify them here.

dashed:
> I think this is more confusing than it needs to be. The interface is leaking business logic.
> 
> It seems like retention settings can specify:
> 
> 1. standard retention (retention for unsampled data)
> 2. downsampled retetion (for donwsampled data)
> 
> from my perspective, a message which exposes those two numbers should be the only thing exposed in the protos. The services which will use these message structures will have the policy decisions specified in their code, there is no need to specify them here.

@volokluev i ended up removing the inherit standard to simplify the protos further.

proto/sentry_protos/billing/v1/common/v1/retention.proto
```diff
@@ -0,0 +1,32 @@
...
+  // policy, a present zero is a legacy compatibility value: the retention
+  // resolver uses the effective standard duration. A positive value is a
+  // concrete downsampled retention duration.
+  optional uint32 downsampled_days = 2;
+}
+
+// A complete concrete retention policy identified by billing data category.
+message DataCategoryRetention {
+  // DATA_CATEGORY_UNSPECIFIED and DATA_CATEGORY_UNKNOWN are invalid.
+  sentry_protos.billing.v1.DataCategory category = 1;
```
volokluev: I would imagine that you want to have this be by lineitem, not datacategory since that's what the platform operates on

volokluev (approved):

proto/sentry_protos/billing/v1/common/v1/retention.proto
dashed: I initially tried to reconcile retentions by line items. but it doesn't work out cleanly since line items can map to different underlying data categories (via billable metrics).

PR #370, opened 2026-07-17
feat(billing): Add contract-owned retention override config

dashed:
Closes https://linear.app/getsentry/issue/REVENG-331/introduce-retention-proto-support-for-contracts-customers-subscription

This PR adds retention overrides (that are specific to the customer) at the contract level. 

There are two retention override levels at the customer level:
- category specific retention overrides (e.g. spans)
- and organization wide retention override.

The analogue for this in the legacy billing system is:
-  `BillingMetricHistory.retention_days` and `BillingMetricHistory.retention_days_downsampled` https://github.com/getsentry/getsentry/blob/b9f4cddb1d463c5f56083cad5b1201f165f52ccd/getsentry/models/billingmetrichistory.py#L491-L493
- `Subscription.retention_days` https://github.com/getsentry/getsentry/blob/60d70fe37a7d8431842357aba588e274925fb430/getsentry/models/subscription.py#L256

### DIFF
--- a/proto/sentry_protos/billing/v1/services/contract/v1/retention_config.proto
+++ b/proto/sentry_protos/billing/v1/services/contract/v1/retention_config.proto
@@ -1,0 +1,42 @@
+syntax = "proto3";
+
+package sentry_protos.billing.v1.services.contract.v1;
+
+import "sentry_protos/billing/v1/data_category.proto";
+
+// A single customer's sparse departure from the package retention defaults for
+// one billing data category.
+message RetentionOverride {
+  // The billing data category this override applies to. Package defaults and
+  // contract overrides use the same billing category identity.
+  sentry_protos.billing.v1.DataCategory category = 1;
+
+  // Standard/full-fidelity retention in calendar days. Absence means no
+  // standard override; a present value must be positive.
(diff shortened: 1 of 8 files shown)


proto/sentry_protos/billing/v1/services/contract/v1/retention_config.proto
```diff
@@ -0,0 +1,45 @@
...
+// The contract-owned retention overrides. A service layers these overrides on
+// top of the package defaults.
+//
+// This message is raw sparse input to effective-retention resolution. It is not
+// the resolved policy. If a field is absent, that field falls through to its
+// package retention chain.
+message RetentionConfig {
+  // The identifier of the retention policy revision this config projects. This
+  // identifier is opaque. Do not parse it.
+  string revision = 1;
```
brendanhsentry: do we need this?

dashed: @brendanhsentry not anymore. this was previously for retention db tables that are now replaced with their user parameter counter parts.

noahsmartin (approved):

PR #374, opened 2026-07-20
feat(billing-platform): Immediate contract-change protos for checkout

skonves:
Protos for the checkout immediate tier/interval change flow:

- **`ChangeContractResponse.invoice_id`** — the prorated invoice a committed immediate change creates (unset when deferred or no-charge).
- **`RolloverContractRequest.immediate`** — roll a contract over now (mid-term) instead of at period end, applying the request's change to the new contract.
- **`BillContractChange`** (new invoicer endpoint) — invoice an immediate mid-term change and roll over now, given the change to apply + prorated line items.

Replaces the earlier `ApplyImmediateContractChange` endpoint (dropped): the immediate change is now composed from the invoicer + rollover, not a bespoke contract-service path.


proto/sentry_protos/billing/v1/services/contract/v1/endpoint_apply_immediate_contract_change.proto
```diff
@@ -0,0 +1,34 @@
...
+message ApplyImmediateContractChangeRequest {
+  uint64 contract_id = 1;
+
+  // The package the contract switches to. If unset, the package is unchanged.
+  optional string package_uid = 2;
+
+  // The month_interval the contract switches to. If unset, it is unchanged. A
+  // change here (e.g. monthly -> annual) starts a fresh term now rather than
+  // keeping the current term.
+  optional uint32 month_interval = 3;
```
volokluev: why is it a uint32? what does the number actually represent?

skonves: This is the `1` or `12` that we've been using to indicate "monthly" vs "annual". It's the same nomenclature that shows up in the Contract and elsewhere.

noahsmartin: we made it an int so that we could change it to arbitrary larger durations in the future. Like a 24 month contract etc... Emerge had some of these but Sentry doesn't (yet)

proto/sentry_protos/billing/v1/services/contract/v1/endpoint_apply_immediate_contract_change.proto
```diff
@@ -0,0 +1,34 @@
...
+
+  // The prorated invoice's line items (the new plan's charge plus a credit for
+  // the unused portion of the current plan), computed by the caller.
+  repeated InvoiceLineItem line_items = 4;
+
+  // User configs (PAYG budgets and/or reservation overrides) to apply. If
+  // empty, the contract's user configs are unchanged.
+  repeated UserConfig user_configs = 5;
+}
```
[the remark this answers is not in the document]
noahsmartin: hmm if this is creating an invoice, this comment might be accurate

noahsmartin (approved):

PR #377, opened 2026-07-21
ref: Add line item types for invoice analytics

noahsmartin:
The item_type string is required used in the line items table by downstream analytics pipelines. We can serve it from platform data so that launching a new SKU only requires changes in the package service.

### DIFF
--- a/proto/sentry_protos/billing/v1/common/v1/line_item_details.proto
+++ b/proto/sentry_protos/billing/v1/common/v1/line_item_details.proto
@@ -5,6 +5,16 @@ package sentry_protos.billing.v1.common.v1;
 import "sentry_protos/billing/v1/common/v1/billable_metric.proto";
 import "sentry_protos/billing/v1/common/v1/unit_info.proto";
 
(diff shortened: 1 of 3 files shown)


armenzg (approved):

brendanhsentry (approved):

proto/sentry_protos/billing/v1/common/v1/line_item_details.proto
```diff
@@ -5,6 +5,16 @@ package sentry_protos.billing.v1.common.v1;
 import "sentry_protos/billing/v1/common/v1/billable_metric.proto";
 import "sentry_protos/billing/v1/common/v1/unit_info.proto";
 
+// The invoice line item type strings used when charging for a line item,
+// distinguished by how the charge was incurred.
+message InvoiceData {
```
volokluev: Reading the code implementation PR, I think this is a confusing name. I assume "InvoiceData" would be all the data of an invoice. This is more like "InvoiceMetadataTags"?

PR #379, opened 2026-07-23
feat(billing): Add BustQuotasCache endpoint protos

armenzg:
Adds `BustQuotasCacheRequest` / `BustQuotasCacheResponse` under the quotas service so GetSentry can invalidate the contract-scoped Django cache used by `GetQuotas` through a proper protobuf service boundary.

Callers that also schedule Relay project-config refresh can treat cache bust as best-effort: a missing current contract or lookup failure returns an empty response rather than failing the refresh path.

This supports this PR: https://github.com/getsentry/getsentry/pull/21175


proto/sentry_protos/billing/v1/services/quotas/v1/endpoint_bust_quotas_cache.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto3";
+
+package sentry_protos.billing.v1.services.quotas.v1;
+
+// Invalidates the contract-scoped Django cache entry used by GetQuotas for the
```
noahsmartin, line 5: I would think of the "contract-scoped Django cache entry" as an implementation detail that doesn't belong in the interface definition here. The interface only has organization_id, so we might not want to mention contract in this comment

noahsmartin (approved):

PR #383, opened 2026-07-24
chore(billing-platform): Add can_product_trial to line items

brendanhsentry:
will be used to address [this TODO](https://github.com/getsentry/getsentry/blob/1f4e3a405dbe04cf155a119043b84aaed81f9da2/getsentry/billing/platform/services/engagement/service.py#L537-L538)

### DIFF
--- a/proto/sentry_protos/billing/v1/services/package/v1/package.proto
+++ b/proto/sentry_protos/billing/v1/services/package/v1/package.proto
@@ -34,6 +34,9 @@ message LineItemConfig {
     bool is_unlimited_trial = 10;
     uint64 num_trial_units = 11;
   }
+
+  // Whether this LineItem is eligible for a product trial.
+  bool can_product_trial = 12;
 }
 
 // Represents a budget included in a package that is collectively used by one or more line items,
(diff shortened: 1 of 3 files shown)


proto/sentry_protos/billing/v1/services/package/v1/package.proto
```diff
@@ -34,6 +34,9 @@ message LineItemConfig {
     bool is_unlimited_trial = 10;
     uint64 num_trial_units = 11;
   }
+
+  // Whether this LineItem is eligible for a product trial.
+  bool can_product_trial = 12;
```
skonves, line 39: I wonder if this would be better represented as an "acquisition mode" in the package catalog.

brendanhsentry: I'm not sure if that would work. Acquisition mode is a property of a package and this needs to be a property of line items. We need to specify the fact that errors, crons, and uptime aren't eligible for product trials.

skonves (approved):

noahsmartin (approved):

PR #384, opened 2026-07-27
feat(billing): Add has_remaining_capacity

armenzg:
## Summary
- Adds `has_remaining_capacity` on `LineItemUsageSummary` (field 8) and a pool-level mirror on `SharedLineItemUsageSummary` (field 5) so quota enforcement can detect remaining Engagement unit grant capacity without changing `reserved_budget_exhausted` / `payg_budget_exhausted` semantics.

**Follow-up:** getsentry bumps this package and uses both fields for UsagePricer + ContractBudgetEnforcer + `get_unit_grants` windows (https://github.com/getsentry/getsentry/pull/21233).

## Test plan
- [x] Buf lint / format / breaking checks
- [ ] getsentry consumes regenerated bindings and exercises mid-period grant windows


proto/sentry_protos/billing/v1/services/engagement/v1/recurring_credit.proto
```diff
@@ -32,4 +34,8 @@ message RecurringCredit {
   optional string line_item_uid = 1;
   RecurringCreditType type = 2;
   uint64 amount = 3;
+  // UTC calendar day the credit row was created. Unit-grant consumers use
+  // max(ondemand_period_start, date_added) so mid-period grants only drain
+  // against usage on/after the grant day (forward headroom).
```
noahsmartin: @brendanhsentry what do you think of this? I thought you were already handling this case (grant added mid period) it doesn't seem like this date should need to be exposed since this already is taken into account

proto/sentry_protos/billing/v1/services/usage_pricer/v1/endpoint_usage_pricer.proto
```diff
@@ -71,6 +71,12 @@ message LineItemUsageSummary {
   // period. Only set for members of a shared reserved-budget pool (e.g. Seer);
   // 0 for line items without a reserved budget.
   uint64 reserved_spend_cents = 7;
+
+  // True when Engagement unit grants (including trial-sourced grants) for this
+  // line item or its shared pool still have remaining units/cents after draining
+  // priced usage. Unlimited active grants are true. Independent of reserved/PAYG
+  // exhaustion; used by quota enforcement to unblock ingest after mid-period grants.
+  bool has_remaining_granted_capacity = 8;
```
noahsmartin: I think it would be best if this was independent of grants. The API from the usage pricer could just be `has_remaining_capacity: bool` This would allow callers of the usage pricer to just know if ingest needs to be blocked or not, but does not tie the implementation specifically to grants. I think we want to hide the fact that right now grants are the only mechanism to increase capacity as an implementation detail and not make it part of the interface

proto/sentry_protos/billing/v1/services/engagement/v1/recurring_credit.proto
brendanhsentry: Yeah I don't think we need this. I don't see how it would differ from `EffectiveUnitGrant.start_date`

armenzg: I have removed this from here and the getsentry PR.

proto/sentry_protos/billing/v1/services/usage_pricer/v1/endpoint_usage_pricer.proto
armenzg: I made it independent.

noahsmartin (approved):

PR #392, opened 2026-08-03
chore(billing-platform): Add start_new_term to from checkout service to rollover

brendanhsentry:
Checkout service determines whether we should apply immediate changes to the current term or if we should start a new billing period.

### DIFF
--- a/proto/sentry_protos/billing/v1/services/contract/v1/endpoint_rollover_contract.proto
+++ b/proto/sentry_protos/billing/v1/services/contract/v1/endpoint_rollover_contract.proto
@@ -26,12 +26,7 @@ message RolloverContractRequest {
   optional string tax_transaction_code = 7;
 
   // Roll the contract over as of now instead of at its period end: the current
-  // contract is closed immediately and the new contract starts now. Used for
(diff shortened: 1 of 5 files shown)


proto/sentry_protos/billing/v1/services/contract/v1/endpoint_rollover_contract.proto
```diff
@@ -33,6 +33,9 @@ message RolloverContractRequest {
   // interval stays co-terminous with the current term, while an interval
   // change (or an already-ended term) starts a fresh term now.
   bool immediate = 8;
+
+  // Whether the requested change should result in a new billing period.
+  bool start_new_term = 9;
```
skonves, line 33: We might want to combine this with `immediate` in some form of enum. As it stands, both `immediate` and `start_new_term` being set to `true` seems to be self-contradictory.

skonves (approved):
We should update the comment on `immediate` to just

```
  // Roll the contract over as of now instead of at its period end: the current
  // contract is closed immediately and the new contract starts now.
```

I think the rest is now either wrong or misleading.

PR #393, opened 2026-08-03
feat(snuba): Add FUNCTION_COLLECT_UNIQUE for array-returning aggregations

pbhandari:
## Summary

Adds `FUNCTION_COLLECT_UNIQUE = 14` to the snuba `Function` enum in `trace_item_attribute.proto`. Unlike the existing aggregations that reduce a group to a scalar, this one collects each group's distinct values into an array (backed by ClickHouse `groupUniqArray`) and is returned as a `val_array` on `TraceItemTable`.

Includes the regenerated Rust bindings and `Cargo.lock` bump per the repo's codegen workflow.

## Related

- Linear: [EAP-623 — Array aggregation support for GetTraceItemTable](https://linear.app/getsentry/issue/EAP-623/array-aggregation-support-for-gettraceitemtable)
- Consumer PR (snuba, implements the aggregation against this enum value): getsentry/snuba#8256


proto/sentry_protos/snuba/v1/trace_item_attribute.proto
```diff
@@ -138,6 +138,7 @@ enum Function {
   FUNCTION_MIN = 10;
   FUNCTION_UNIQ = 11;
   FUNCTION_ANY = 13;
+  FUNCTION_COLLECT_UNIQUE = 14;
```
onewland, line 141: could we not repurpose FUNCTION_UNIQ here? what is the difference between the two semantically?

pbhandari: FUNCTION_UNIQ uses clickhouse's [uniq](https://clickhouse.com/docs/reference/functions/aggregate-functions/uniq) which is basically a `dedup + count`.

This new function is going to use the [groupUniqArray](https://clickhouse.com/docs/reference/functions/aggregate-functions/groupUniqArray#groupUniqArray) which is just going to `dedup + return`.

I think the behaviour is different enough that we don't want to repurpose that function.

onewland: sounds good

